PR #73, opened 2025-12-09
[WIP] feat: populate wp status

### LINKED ISSUE #41: [time boxed] Expose the policies status inside our WorkloadPolicy resource
Today, we don't provide any feedback to our users on whether the WorkloadPolicy correctly deploys the underlying Tetragon TracingPolicy on the system.

Tetragon may fail to apply our policy on some nodes, but we have no feedback. 
To check that everything is in place, the user has 2 choices:

 - look at tetragon Prometheus metrics
 - look at tetragon logs

This is not just difficult for our users, but our own controllers also need to know when a policy is correctly deployed to delete the previous one (see here for more details https://github.com/neuvector/runtime-enforcer/issues/5)

**Note**: after @Andreagit97's bespoke implementation of a runtime enforcement agent based on our own code the text above became a little bit obsolete. We're anyway talking about implementing a status field into the `WorkloadPolicy` resource.

Andreagit97:

This PR implements what is described in #41:

* We expose a gRPC API from our daemonset. This API is really simple at the beginning, just a method ListPolicies that tells us the state of the policy and the name (we can enrich it in the future)      
* We create a new controller that regularly scrapes these gRPC endpoints (it keeps the gRPC connection open and checks that everything is still up)
* It aggregates the state for all the policies on all the nodes
* It iterates over the WPs and checks if we need to update the status for some of them

Issue #41 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


holyspectral (commented):
We've discussed this before so yeah it looks good to me so far.  Just some high level comments.

PR #222, opened 2026-01-28
feat: introduce grpc server inside the agent and a WP status controller to update the resource status

### LINKED ISSUE #175: [time-boxed] create a GRPC API to expose WorkloadPolicy status from each agent and use it in the WorkloadPolicy controller
The scope of this issue is to create a GRPC API to expose the policy status from each agent and to consume this API inside the WorkloadPolicy controller, to populate the status of the WorkloadPolicy CR.

Acceptance criteria:
- [ ] Create a GRPC API to expose the policy status from each agent. A GRPC API to do that with Tetragon was already implemented here https://github.com/neuvector/runtime-enforcer/pull/73, we can probably slightly adapt it.
- [ ] Use the GRPC API inside the WP controller to aggregate and populate the status of WP.  https://github.com/neuvector/runtime-enforcer/pull/73 implements a model in which the WorkloadPolicy controller reconciles every `n` seconds and scrapes the gRPC API from all agents. This was the result of the discussion here https://github.com/neuvector/runtime-enforcer/issues/41, so I believe we can follow this approach.

Andreagit97:

This PR does 2 main things:
-  adds a grpc server inside the agent so that consumers can get the policy status.
  - This PR adds only the grpc channel and a mock method to get the state, but it doesn't populate the policy state inside the agent. This should be tackled in another issue https://github.com/neuvector/runtime-enforcer/issues/174
- adds a controller that gets the mock state from each agent and populates the workload policy status. Today, the controller always receives a mock state from each node (POLICY_STATE_READY).

This PR revisits an old one https://github.com/neuvector/runtime-enforcer/pull/73

fixes #175 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/templates/operator/deployment.yaml
+++ b/charts/runtime-enforcer/templates/operator/deployment.yaml
@@ -27,7 +27,6 @@ spec:
       - args: 
         - --wp-status-reconciler-agent-grpc-port={{ .Values.agent.agent.grpcExporterPort }}
         - --wp-status-reconciler-update-interval={{ .Values.operator.manager.wpStatusUpdateInterval }}
-        - --wp-status-reconciler-agent-namespace={{ .Release.Namespace }}
         - --wp-status-reconciler-agent-label-selector={{ include "runtime-enforcer.agent.labelSelectorString" . }}
         - --wp-status-reconciler-agent-grpc-mtls-cert-dir={{ include "runtime-enforcer.grpc.certDir" . }}
         {{- toYaml .Values.operator.manager.args | nindent 8 }}
--- a/internal/controller/workloadpolicystatus_sync.go
+++ b/internal/controller/workloadpolicystatus_sync.go
@@ -63,10 +63,6 @@ func NewWorkloadPolicyStatusSync(
 		return nil, fmt.Errorf("invalid update interval: %v", config.UpdateInterval)
 	}
 
-	if config.AgentNamespace == "" {
-		return nil, errors.New("empty agent namespace")
-	}
-
 	agentLabelSelector := make(map[string]string)
 	labels := strings.Split(config.AgentLabelSelector, ",")
 	for _, label := range labels {
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/tls"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -29,6 +30,10 @@ import (
 	securityv1alpha1 "github.com/rancher-sandbox/runtime-enforcer/api/v1alpha1"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/controller"
 	// +kubebuilder:scaffold:imports
+)
+
+const (
+	namespaceNamePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 )
 
 type Config struct {
@@ -71,10 +76,6 @@ func parseArgs(logger logr.Logger, config *Config) {
 		"wp-status-reconciler-update-interval",
 		0,
 		"The interval at which the workload policy status reconciler updates the status of WorkloadPolicy resources.")
-	flag.StringVar(&config.wpStatusSyncConfig.AgentNamespace,
-		"wp-status-reconciler-agent-namespace",
-		"",
-		"The namespace where the agent pods are deployed.")
 	flag.StringVar(&config.wpStatusSyncConfig.AgentLabelSelector,
 		"wp-status-reconciler-agent-label-selector",
 		"",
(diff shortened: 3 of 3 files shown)


internal/controller/workloadpolicystatus_controller.go
```diff
@@ -0,0 +1,208 @@
...
+				&MockEventHandler{},
+			)).Complete(r)
+
+	// Send a first event to kickstart the reconciliation loop. The controller will requeue itself after each reconciliation.
+	// todo!: let's see if we can remove it.
+	go func() {
+		time.Sleep(1 * time.Second)
+		startChan <- event.TypedGenericEvent[MockEvent]{Object: MockEvent{}}
+		close(startChan)
+	}()
```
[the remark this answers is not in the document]
Andreagit97: Great suggestion, I believe a simple runnable is the best fit at the moment. 
Done :)

internal/controller/mockevent.go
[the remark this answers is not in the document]
Andreagit97: since we now use a runnable we don't need this one anymore

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -0,0 +1,199 @@
...
+	issue    v1alpha1.NodeIssue
+	policies map[string]*pb.PolicyStatus
+}
+
+// nodesInfoMap maps node names to their info.
+// Structure: NodeName -> Info.
+type nodesInfoMap map[string]nodeInfo
+
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=security.rancher.io,resources=workloadsecuritypolicies/status,verbs=get;update;patch
```
kyledong-suse:
```suggestion
// +kubebuilder:rbac:groups=security.rancher.io,resources=workloadpolicies/status,verbs=get;update;patch
```

charts/runtime-enforcer/templates/operator/role.yaml
```diff
@@ -53,6 +53,7 @@ rules:
   resources:
   - workloadpolicies/status
   - workloadpolicyproposals/status
+  - workloadsecuritypolicies/status
```
kyledong-suse: We don't have `workloadsecuritypolicies` anymore, so I think it should be removed.

internal/grpcexporter/server.go
```diff
@@ -0,0 +1,39 @@
...
+	lc := net.ListenConfig{}
+	addr := fmt.Sprintf(":%d", s.port)
+	listener, err := lc.Listen(ctx, "tcp", addr)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %s: %w", addr, err)
+	}
+	grpcServer := grpc.NewServer()
+	pb.RegisterAgentObserverServer(grpcServer, newAgentObserver(s.logger, s.resolver))
+	s.logger.InfoContext(ctx, "Starting gRPC exporter", "addr", addr)
+	return grpcServer.Serve(listener)
```
kyledong-suse: The gRPC library supports this with [GracefulStop](https://grpc.io/docs/guides/server-graceful-stop/). I think we should take advantage of this method, so the agent can shut down the gRPC server properly.

internal/controller/workloadpolicystatus_test.go
```diff
@@ -0,0 +1,231 @@
+//nolint:testpackage //we want to test unexported functions
```
dottorblaster: Maybe in the future we can think about having a naming convention for tests that can "escape" this linter rule.

charts/runtime-enforcer/templates/_helpers.tpl
```diff
@@ -60,3 +60,18 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Grpc mTLS helpers
+*/}}
+{{- define "runtime-enforcer.grpc.CertDir" -}}
```
holyspectral: It's super minor, but I think helm charts uses camelCase. 

```suggestion
{{- define "runtime-enforcer.grpc.certDir" -}}
```

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+			status.AddTransitioningNode(nodeName)
+		case pb.PolicyState_POLICY_STATE_ERROR:
+			status.AddNodeIssue(nodeName, v1alpha1.NodeIssue{
+				Code: v1alpha1.NodeIssuePolicyFailed,
+				// todo!: we should receive the error message from the agent.
+				Message: "policy is in error state",
+			})
+		case pb.PolicyState_POLICY_STATE_UNSPECIFIED:
+		default:
+			panic(fmt.Sprintf("unhandled tracing policy state: %v", policyStatus.GetState()))
```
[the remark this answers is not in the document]
Andreagit97: I have to admit I'm a big fan of panics in these cases, but yes I can turn them into errors if you prefer

charts/runtime-enforcer/templates/_helpers.tpl
Andreagit97: ty!

charts/runtime-enforcer/templates/operator/deployment.yaml
```diff
@@ -24,7 +24,13 @@ spec:
       {{- toYaml .Values.imagePullSecrets | nindent 8 }}
       {{- end }}
       containers:
-      - args: {{- toYaml .Values.operator.manager.args | nindent 8 }}
+      - args: 
+        - --wp-status-reconciler-agent-grpc-port={{ .Values.agent.agent.grpcExporterPort }}
+        - --wp-status-reconciler-update-interval={{ .Values.operator.manager.wpStatusUpdateInterval }}
+        - --wp-status-reconciler-agent-namespace={{ .Release.Namespace }}
```
holyspectral: minor: we can get the namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file inside the operator, so we don't have to allow this to be configurable.

PR #223, opened 2026-01-28
feat: support container add/remove in WorkloadPolicy

### LINKED ISSUE #209: Feature Request: allow users to update a WorkloadPolicy adding/removing containers
### Is your feature request related to a problem?

_No response_

### Solution you'd like

Let's consider an example to better understand the feature:

- WorkloadPolicy has rules for container `main`.
- Pod definition has two containers: `main` and `sidecar`. The pod has the label configured to consume our policy
- Pod starts and is bound to the WorkloadPolicy. The `main` container has limitations in place, while the `sidecar` one has none

The user changes the `WorkloadPolicy`:
- New allowed programs are added to `main`
- a new set of rules is added to the `sidecar` container

Our current agent implementation:
- updates the bpf map that stores the list of programs allowed inside of the `main` container
- does nothing for the `sidecar` container

With this feature, our agent should react to the addition of the `sidecar` container in the WP and should start to protect it. We should probably support container deletion as well

### Alternatives you've considered

_No response_

kyledong-suse:

support per-container WorkloadPolicy add/remove with runtime updates

fixes #209 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/resolver/policy.go
+++ b/internal/resolver/policy.go
@@ -127,28 +127,31 @@ func (r *Resolver) applyPolicyToPodIfPresent(state *podState) error {
 
 // syncWorkloadPolicy ensures state and BPF maps match wp.Spec.RulesByContainer:
 // allocates a policy ID for new containers, (re)applies binaries and mode for every container in the spec.
+// It returns the container→policyID map for newly created policy IDs.
 // This must be called with the resolver lock held.
-func (r *Resolver) syncWorkloadPolicy(wp *v1alpha1.WorkloadPolicy) error {
+func (r *Resolver) syncWorkloadPolicy(wp *v1alpha1.WorkloadPolicy) (policyByContainer, error) {
 	wpKey := wp.NamespacedName()
 	mode := policymode.ParseMode(wp.Spec.Mode)
 	state := r.wpState[wpKey]
+	newContainers := make(policyByContainer)
 
 	for containerName, containerRules := range wp.Spec.RulesByContainer {
 		polID, hadPolicyID := state[containerName]
 		op := bpf.ReplaceValuesInPolicy
 		if !hadPolicyID {
 			polID = r.allocPolicyID()
-			state[containerName] = polID
+			newContainers[containerName] = polID
 			r.logger.Info("create policy", "id", polID,
 				"wp", wpKey,
 				"container", containerName)
 			op = bpf.AddValuesToPolicy
 		}
 		if err := r.upsertPolicy(polID, containerRules.Executables.Allowed, mode, op); err != nil {
-			return fmt.Errorf("failed to populate policy for wp %s, container %s: %w", wpKey, containerName, err)
+			return nil, fmt.Errorf("failed to populate policy for wp %s, container %s: %w", wpKey, containerName, err)
 		}
 	}
-	return nil
+
+	return newContainers, nil
 }
 
 // handleWPAdd adds a new workload policy into the resolver cache and applies the policies to all running pods that require it.
@@ -166,20 +169,24 @@ func (r *Resolver) handleWPAdd(wp *v1alpha1.WorkloadPolicy) error {
 		return fmt.Errorf("workload policy already exists in internal state: %s", wpKey)
 	}
 
-	r.wpState[wpKey] = make(policyByContainer, len(wp.Spec.RulesByContainer))
-
-	if err := r.syncWorkloadPolicy(wp); err != nil {
+	state := make(policyByContainer, len(wp.Spec.RulesByContainer))
+	r.wpState[wpKey] = state
+	var err error
+	var newContainers policyByContainer
+	if newContainers, err = r.syncWorkloadPolicy(wp); err != nil {
 		return err
 	}
+	for containerName, policyID := range newContainers {
+		state[containerName] = policyID
+	}
 
-	wpMap := r.wpState[wpKey]
 	// Now we search for pods that match the policy
 	for _, podState := range r.podCache {
 		if !podState.matchPolicy(wp.Name) {
 			continue
 		}
 
-		if err := r.applyPolicyToPod(podState, wpMap); err != nil {
+		if err = r.applyPolicyToPod(podState, state); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


Andreagit97:
Thank you for the rework! IMO, we can factorize even more, reusing the same applyPolicyToPod method without looping again into all pods. I've crafted more or less the changes https://github.com/Andreagit97/runtime-enforcer/pull/1 (not tested, but should give you the idea) WDYT?

I've also added a unit test for the resolver in that PR, maybe it's a good time to start testing it with unit tests

internal/resolver/policy.go
```diff
@@ -121,65 +137,80 @@ func (r *Resolver) handleWPAdd(wp *v1alpha1.WorkloadPolicy) error {
...
 
-		// Atomically replace values in BPF maps
-		if err := r.policyUpdateBinariesFunc(policyID, newRules.Executables.Allowed, bpf.ReplaceValuesInPolicy); err != nil {
-			return fmt.Errorf("failed to replace policy values for wp %s, container %s: %w",
+	// Clean up removed containers: cgroups, policy values, mode, and state.
+	for containerName := range removedContainers {
+		policyID := state[containerName]
+		cgroupIDs := r.getCgroupIDsForContainer(wp.Name, containerName)
+		if len(cgroupIDs) > 0 {
+			if err := r.cgroupToPolicyMapUpdateFunc(PolicyIDNone, cgroupIDs, bpf.RemoveCgroups); err != nil {
```
flavio: I was initially confused about why we have to pass `PolicyIDNone`, given we have `policyID`. I think the policy ID doesn't matter when a removal operation is done; hence we don't care about what we pass.
Am I right?

I already had the suspect the signatures/names of these helper functions were a bit misleading. Something to improve in the future IMHO.

internal/resolver/policy.go
Andreagit97: IMO we need to remove this one, because now we are providing `applyPolicyToPod` with a diff, so it's possible that some containers are not present in `policyByContainer`, but they are already protected, see the WPUpdate case, where we provide the `changes` map.
We could add a comment to explain it

internal/resolver/policy.go
```diff
@@ -121,65 +169,49 @@ func (r *Resolver) handleWPAdd(wp *v1alpha1.WorkloadPolicy) error {
...
-			"old-count", len(oldRules.Executables.Allowed),
-			"new-count", len(newRules.Executables.Allowed),
-		)
-
-		// Atomically replace values in BPF maps
-		if err := r.policyUpdateBinariesFunc(policyID, newRules.Executables.Allowed, bpf.ReplaceValuesInPolicy); err != nil {
-			return fmt.Errorf("failed to replace policy values for wp %s, container %s: %w",
-				wpKey, containerName, err)
+		policyID := state[containerName]
+		if err := r.syncPolicyInBPF(policyID, nil, 0, bpf.RemoveValuesFromPolicy); err != nil {
```
Andreagit97: I think this is ok, but I would leave a comment here because it could open up to dangerous situations. At this point, we still have the association cgroup -> policy_id, so removing the mode and binary values, we are allowing all binaries.... the flow is the following:

- We don't find any binaries associated with the policyID
- We look for the mode value to send the notification to userspace, but we don't find it https://github.com/rancher-sandbox/runtime-enforcer/blob/5d8a20d5d330dd4bce63d63df4343780a43ef93f/bpf/main.c#L569
- We don't enforce the policy, and we let the binary to run.

We should also remove the log error ebpf side https://github.com/rancher-sandbox/runtime-enforcer/blob/96b5dd68b560a4e2c33db167767eaaa554c3444b/bpf/main.c#L568 and explain with a comment that this situation could happen during policy removals

internal/resolver/policy_test.go
```diff
@@ -0,0 +1,101 @@
...
+	state := r.wpState[key]
+	require.Len(t, state, 2)
+	require.Contains(t, state, "c1")
+	require.Contains(t, state, "c2")
+	ids := make(map[PolicyID]struct{})
+	for _, id := range state {
+		ids[id] = struct{}{}
+	}
+	require.Equal(t, map[PolicyID]struct{}{PolicyID(1): {}, PolicyID(2): {}}, ids)
+	initialState := r.wpState[key]
```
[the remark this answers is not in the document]
Andreagit97: Good point, thanks for the explanation

internal/resolver/policy.go
```diff
@@ -68,6 +89,36 @@ func (r *Resolver) applyPolicyToPodIfPresent(state *podState) error {
 	return r.applyPolicyToPod(state, pol)
 }
 
+// syncWorkloadPolicy ensures state and BPF maps match wp.Spec.RulesByContainer:
+// allocates a policy ID for new containers, (re)applies binaries and mode for every container in the spec.
+// If changes is non-nil, we will use it to track changes for the update operation.
+// This must be called with the resolver lock held.
+func (r *Resolver) syncWorkloadPolicy(wp *v1alpha1.WorkloadPolicy, changes policyByContainer) error {
```
flavio: Changes is used to store the list of new container -> policyIDs that might be created.

I propose to change the function signature to be:

```
func (r *Resolver) syncWorkloadPolicy(wp *v1alpha1.WorkloadPolicy) (policyByContainer, error)
```

We are now always returning the map that holds the new policy IDs that have been created.

Because of that, I would also not use the name `changes` inside of the function to refer to this variable.

This makes explicit what is going on and doesn't have the caller of function wonder whether they should pass a value for `changes` or not. They can simply ignore the returned value if they don't care about that. This feels more idiomatic to me.

flavio (changes requested):

kyledong-suse:
I'm still working on fixing the e2e test, so please don't review it for now. Thanks.

holyspectral (commented):
Sorry for being late.  I think the overall code is already in a good shape.  I just have some general comments.

internal/resolver/policy.go
```diff
@@ -68,6 +125,32 @@ func (r *Resolver) applyPolicyToPodIfPresent(state *podState) error {
...
+	wpKey := wp.NamespacedName()
+	mode := policymode.ParseMode(wp.Spec.Mode)
+	state := r.wpState[wpKey]
+
+	for containerName, containerRules := range wp.Spec.RulesByContainer {
+		polID, hadPolicyID := state[containerName]
+		op := bpf.ReplaceValuesInPolicy
+		if !hadPolicyID {
+			polID = r.allocPolicyID()
+			state[containerName] = polID
```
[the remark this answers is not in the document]
kyledong-suse: That's a good suggestion, I'll return the new added containers and update the cache after this reconcile logic call.

holyspectral (approved):

PR #226, opened 2026-01-29
feat: make agent protection persistent during rolling update

### LINKED ISSUE #96: [time boxed] Guarantee protection during agent updates
Today even if we pin ebpf programs/maps in a container on a host bpffs volume, we can't guarantee no downtime because ebpf programs/maps and the protocol with userspace may change across versions.  

This case is to define a strategy for keeping protection as much as possible during upgrade. This problem was discussed on [this issue](https://github.com/neuvector/runtime-enforcer/issues/95). There we decided to experiment with DaemonSet rollouts.

# Admission Criteria

- [x] Change the deployment strategy of the agent DaemonSet to use rollout. In this way, the old agent will be stopped only when the new agent is up and running.
- [x] Ensure no conflicts happen while the two agents are running at the same time.

holyspectral:

This PR introduces a updateStrategy and a readinessProbe, which ensures that during a rolling update, the old agent will not be terminated before the new agent is ready.

A new e2e test is added to simulate a `kubectl rollout restart`, but it's not meant to test helm upgrade.  A separate test would still be necessary.

fixes #96 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/templates/agent/daemonset.yaml
+++ b/charts/runtime-enforcer/templates/agent/daemonset.yaml
@@ -69,7 +69,7 @@ spec:
             path: /readyz
             port: 8081
           initialDelaySeconds: 5
-          periodSeconds: 10
+          periodSeconds: 3
         resources: {{- toYaml .Values.agent.agent.resources | nindent 10 }}
         securityContext: {{- toYaml .Values.agent.agent.containerSecurityContext | nindent 10 }}
         volumeMounts:
--- a/test/e2e/rolling_update_test.go
+++ b/test/e2e/rolling_update_test.go
@@ -12,15 +12,28 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachinerywait "k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/e2e-framework/klient/decoder"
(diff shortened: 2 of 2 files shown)


kyledong-suse (commented):
LGTM, thanks. Just one more comment, do we want to add a test that checks `updateStrategy` and `readinessProbe` in daemonset_test.yaml?

internal/nri/handler.go
```diff
@@ -73,9 +75,14 @@ func (h *Handler) checkNRISupport() error {
 func (h *Handler) startNRIPlugin(ctx context.Context) error {
 	var err error
 
+	defer func() {
+		h.synchronized = false
+	}()
```
flavio: From my understanding, this is useful only when the plugin is restarted because of some communication hiccup with the NRI socket.
I'm fine leaving it around, but why do we need to wrap that into a `defer`?

internal/nri/handler.go
```diff
@@ -18,10 +19,11 @@ const (
...
 type Handler struct {
-	socketPath  string
-	pluginIndex string
-	logger      *slog.Logger
-	resolver    *resolver.Resolver
+	socketPath   string
+	pluginIndex  string
+	logger       *slog.Logger
+	resolver     *resolver.Resolver
+	synchronized bool
```
flavio: Probably not super necessary, but we might use an `atomic.Bool` since there's some concurrency between the code that sets this value and the readiness http check

Andreagit97 (commented):
Thank you, some minor suggestions

PR #245, opened 2026-02-06
fix: safe shutdown of bpf manager

### LINKED ISSUE #112: Improve ebpf map lifecycle management
The eBPF manager provides update callbacks to other components (eventscraper, resolver) in order to modify ebpf maps. If the eBPF manager is stopped (ctx is closed), it will close the eBPF maps, so it is possible that during the turndown phase, an update function is called on a closed map. It is very likely that this will cause a panic. 
We need to investigate if this is really the case, and if so, what is the best way to prevent this situation. The trivial solution is to use a mutex, but this means that for each map update, we need to take a lock...

Some references to the issue:
* https://github.com/neuvector/runtime-enforcer/blob/eef0cf1638a6e5f737abcbbeb63538cbe8469a14/internal/bpf/cgroup_policy.go#L96
* https://github.com/neuvector/runtime-enforcer/blob/eef0cf1638a6e5f737abcbbeb63538cbe8469a14/internal/bpf/cgtracker.go#L23

holyspectral:

1. Add a flag to indicate bpf manager is already shutdown.
2. Make bpf manager's functions return nil error if the bpf manager is already shutdown.

fixes #112 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/bpf/manager.go
+++ b/internal/bpf/manager.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"sync/atomic"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
@@ -51,6 +52,7 @@ type Manager struct {
 	logger           *slog.Logger
 	objs             *bpfObjects
 	policyStringMaps []*ebpf.Map
+	isShuttingDown   atomic.Bool
 
 	// Learning
 	enableLearning    bool
@@ -195,9 +197,21 @@ func (m *Manager) isKernelPre5_9() bool {
 	return m.isPre5_9
 }
 
+func (m *Manager) handleErrOnShutdown(err error) error {
+	// We have multiple go routines to update ebpf maps, e.g., policy informer and NRI plugin.
+	// Because of this, we could receive errors during shutdown flow, e.g., bad file descriptor.
+	// Since we are shutting down, we don't have much to do with these errors, so we just ignore them.
+	if m.isShuttingDown.Load() {
+		return nil
+	}
+	return err
+}
+
 func (m *Manager) Start(ctx context.Context) error {
 	defer func() {
 		m.logger.InfoContext(ctx, "BPF Manager stopped")
+		m.isShuttingDown.Store(true)
+
 		if err := m.objs.Close(); err != nil {
 			m.logger.ErrorContext(ctx, "failed to close BPF objects", "error", err)
 		}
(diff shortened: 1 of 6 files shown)


internal/bpf/manager_test.go
```diff
@@ -174,3 +174,49 @@ func TestMultiplePolicies(t *testing.T) {
...
+
+func TestManagerShutdown(t *testing.T) {
+	manager, cleanup, err := waitRunningManager(t)
+	require.NoError(t, err, "Failed to start manager")
+	cleanup()
+
+	// create a cgroup for test.  We just reuse the existing cgroup if it exists.
+	cgroupPath := "/sys/fs/cgroup/system.slice/test-manager-shutdown.scope"
+	_ = os.MkdirAll(cgroupPath, 0755)
+	defer os.RemoveAll(cgroupPath)
```
Andreagit97: I believe we can remove this code, see below

```suggestion
```

internal/bpf/manager.go
```diff
@@ -195,9 +197,21 @@ func (m *Manager) isKernelPre5_9() bool {
...
 }
 
+func (m *Manager) handleError(err error) error {
+	if err != nil {
+		if m.IsShuttingDown() {
+			return nil
+		}
+		return err
+	}
+	return nil
```
flavio: This is a bit more polished, still from one of @Andreagit97's suggestion on the review:

```suggestion
	if err == nil || m.IsShuttingDown() {
		return nil
	}
	return err
```

Andreagit97: actually we can even do something like this 
```go
func (m *Manager) handleError(err error) error {
	if m.IsShuttingDown() {
		return nil
	}
	return err
}
```

internal/bpf/policy_mode_operation.go
```diff
@@ -43,13 +43,21 @@ func (m *Manager) deletePolicy(policyID uint64) error {
 
 func (m *Manager) GetPolicyModeUpdateFunc() func(policyID uint64, mode policymode.Mode, op PolicyModeOperation) error {
 	return func(policyID uint64, mode policymode.Mode, op PolicyModeOperation) error {
+		var err error
```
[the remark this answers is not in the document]
Andreagit97: nope i was thinking something that we cannot do in golang :laughing: so I would probably say the best approach was what you did

```go
func (m *Manager) GetPolicyModeUpdateFunc() func(policyID uint64, mode policymode.Mode, op PolicyModeOperation) error {
	return func(policyID uint64, mode policymode.Mode, op PolicyModeOperation) error {
		var err error
		switch op {
		case UpdateMode:
			err = m.updatePolicyMode(policyID, mode)
		case DeleteMode:
			err = m.deletePolicy(policyID)
		default:
			panic("unhandled policy mode")
		}
		return m.handleError(err)
	}
}
```
sorry for the noise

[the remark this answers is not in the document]
holyspectral: To add, after some thoughts, I keep my current code, which chains the `m.handleErrOnShutdown()` and the real function call in one line, like the below, because I figured it's more readable than having to go back to see how `err` is handled.  

```
return m.handleErrOnShutdown(m.updatePolicyMode(policyID, mode))
```

Of course this kind of topic is always subjective, so feel free to let me know if you have different thoughts.

PR #246, opened 2026-02-06
ci: support linter of bpf codes

### LINKED ISSUE #100: Add linter for bpf programs
At the time of writing, there is no formatter or linter applied in ebpf codes written in C.   Because of the freedom of C, this can become a problem when we accept more patches in this area.  

Options include:
- https://github.com/cilium/image-tools/blob/master/images/checkpatch/checkpatch.sh

- https://clang.llvm.org/docs/ClangFormat.html (https://github.com/falcosecurity/libs/blob/d4fb9de6724fff602facbfd8a4c20f85f9f2cb25/Makefile#L48-L49)

## Acceptance Criteria

- [x] Add a CI task that enforces the current pre-commit checks that we already have
- [ ] Research a linter for eBPF and add it to our CI

holyspectral:

It turned out it's pretty easy to add a clang-format formatter to our ci, so here we go. 

An example where the CI detects an issue: https://github.com/holyspectral/runtime-enforcer/actions/runs/21767400879/job/62806596492

image

fixes #100 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/updatecli/updatecli.d/pre-commit.yaml
+++ b/updatecli/updatecli.d/pre-commit.yaml
@@ -1,0 +1,85 @@
+name: Update .pre-commit-config.yaml
+sources:
+  clangformatTag:
+    kind: gittag
+    spec:
+      key: name
+      url: https://github.com/pre-commit/mirrors-clang-format
+      versionFilter:
+        kind: semver
+        pattern: 18.1.*
+  golangcilintTag:
+    kind: gittag
+    spec:
+      key: name
+      url: https://github.com/golangci/golangci-lint/
+      versionFilter:
+        kind: latest
+  clangformatVersion:
+    kind: gittag
+    spec:
+      key: hash
+      url: https://github.com/pre-commit/mirrors-clang-format
+      versionFilter:
+        kind: semver
+        pattern: 18.1.*
+  golangcilintVersion:
+    kind: gittag
+    spec:
+      key: hash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -70,6 +70,17 @@ jobs:
       - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: v2.4.0
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install clang format
```
Andreagit97: Thank you for this! If possible i would use pre-commit also in CI so that the clang version is always aligned with what we use locally. See here for an example on how to do that https://github.com/falcosecurity/libs/blob/c5ea0da85bd4d5253c89c70c4cd5815178dd7ca1/.github/workflows/format.yml#L25 (I think here we can just run `pre-commit run --show-diff-on-failure --color=always --all-files` without uploading the artifact)

The idea is to use pre-commit while developing and to use it also in CI to verify everything is ok. WDYT?

Andreagit97: I got it now, thanks!

PR #252, opened 2026-02-09
feat: store workload policy status in agent

### LINKED ISSUE #174: [time-boxed] store workload policy status inside our agent
Each of our agents should internally store the policy deployment status and expose it through grpc, so that a kubectl plugin or a controller could scrape this data.
Today, we just log in in case of failures, and we don't have a real policy status. 
This issue just focuses on storing the state

Acceptance criteria:
- [ ] store the status for each policy inside the agent
- [ ] in case of policy failure, rollback the policy so that the final state is Success/Fail and not intermediate
- [ ] Check the overlap and return an error in [that case](https://github.com/neuvector/runtime-enforcer/blob/eef0cf1638a6e5f737abcbbeb63538cbe8469a14/internal/bpf/cgroup_policy.go#L32)

kyledong-suse:

1. Store WP status inside agent
2. Expose the WP status through GRPC
3. If the WP deployment failed, rollback the WP
4. Add check to avoid overwriting existing policies

fixes #174 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/resolver/policy.go
+++ b/internal/resolver/policy.go
@@ -207,7 +207,7 @@ func (r *Resolver) handleWPAdd(wp *v1alpha1.WorkloadPolicy) error {
 
 	// Now we search for pods that match the policy
 	for _, podState := range r.podCache {
-		if !podState.matchPolicy(wp.Name) {
+		if !podState.matchPolicy(wp.Name, wp.Namespace) {
 			continue
 		}
 
@@ -266,7 +266,7 @@ func (r *Resolver) handleWPUpdate(wp *v1alpha1.WorkloadPolicy) error {
 	}
 
 	for _, podState := range r.podCache {
-		if !podState.matchPolicy(wp.Name) {
+		if !podState.matchPolicy(wp.Name, wp.Namespace) {
 			continue
 		}
 		if err = r.removePolicyFromPod(wpKey, podState, info.polByContainer, removedMap); err != nil {
--- a/internal/resolver/podstate.go
+++ b/internal/resolver/podstate.go
@@ -7,11 +7,17 @@ type podState struct {
 	containers map[ContainerID]*containerInfo
 }
 
-func (pod *podState) matchPolicy(policyName string) bool {
+func (pod *podState) matchPolicy(policyName, policyNamespace string) bool {
 	v, ok := pod.info.labels[v1alpha1.PolicyLabelKey]
 	if !ok || v != policyName {
 		return false
 	}
+
+	// now we need to check if the pod is in the same namespace of the policy since our policies are namespaced.
+	if pod.info.namespace != policyNamespace {
+		return false
+	}
+
 	return true
 }
 


internal/resolver/policy.go
```diff
@@ -6,15 +6,27 @@ import (
 	"github.com/rancher-sandbox/runtime-enforcer/api/v1alpha1"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/bpf"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/types/policymode"
+	pb "github.com/rancher-sandbox/runtime-enforcer/proto/agent/v1"
```
holyspectral: nit: I'd call it `agentv1` instead of `pb`.

internal/resolver/policy.go
```diff
@@ -170,10 +185,15 @@ func (r *Resolver) handleWPAdd(wp *v1alpha1.WorkloadPolicy) error {
...
 
 	state := make(policyByContainer, len(wp.Spec.RulesByContainer))
-	r.wpState[wpKey] = state
+	r.wpState[wpKey] = WPInfo{polByContainer: state}
 	var err error
 	var newContainers policyByContainer
 	if newContainers, err = r.syncWorkloadPolicy(wp); err != nil {
+		// Rollback: remove any new policy IDs that were created (not in state yet).
+		if rollbackErr := r.tearDownPolicyIDs(wpKey, newContainers); rollbackErr != nil {
+			r.logger.Error("failed to rollback policy", "error", rollbackErr)
```
holyspectral: If you want, you can return the original error as well as the rollback error using [errors.Join()](https://pkg.go.dev/errors#Join)

internal/resolver/policy.go
```diff
@@ -146,8 +158,11 @@ func (r *Resolver) syncWorkloadPolicy(wp *v1alpha1.WorkloadPolicy) (policyByCont
 				"container", containerName)
 			op = bpf.AddValuesToPolicy
 		}
-		if err := r.upsertPolicy(polID, containerRules.Executables.Allowed, mode, op); err != nil {
-			return nil, fmt.Errorf("failed to populate policy for wp %s, container %s: %w", wpKey, containerName, err)
+		if err := r.upsertPolicyIDInBPF(polID, containerRules.Executables.Allowed, mode, op); err != nil {
+			// Return newContainers to allow rollback.
+			return newContainers, fmt.Errorf(
+				"failed to populate policy for wp %s, container %s: %w",
+				wpKey, containerName, err)
```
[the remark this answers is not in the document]
kyledong-suse: Yeah, I agree that's a good idea.

holyspectral (approved):
LGTM.  The only nit is that the last commit shouldn't be a `chore`. lol Other than that it looks great.

internal/bpf/cgroup_policy.go
```diff
@@ -42,8 +42,24 @@ func addPolicyToCgroups(cgToPol *ebpf.Map, targetPolID uint64, cgroupIDs []uint6
 
 	for _, cgID := range cgroupIDs {
 		// todo!: check if we can use batch operations and when they are supported
-		// todo!: put in place some checks to avoid overwriting existing policies?
-		if err := cgToPol.Update(&cgID, &targetPolID, ebpf.UpdateAny); err != nil {
+		var existingPolID uint64
+		err := cgToPol.Lookup(&cgID, &existingPolID)
```
Andreagit97: As we said yesterday, not replacing the policy seems a valid approach for now. 
nit: The only thing I would do is to simplify the flow and reduce the number of syscalls in the "good" scenario.

```go
	err := cgToPol.Update(&cgID, &targetPolID, ebpf.UpdateNoExist)
	if err == nil {
		continue
	}
	if !errors.Is(err, ebpf.ErrKeyExist) {
		return fmt.Errorf("failed to update cgroup %d: %w", cgID, err)
	}
    ... and then do the lookup
```

This should reduce the number of syscalls since I expect most of the time the policy is not there, and should also reduce the `if` nesting that is always difficult to follow. WDYT?

internal/resolver/policy.go
```diff
@@ -27,9 +40,9 @@ func (r *Resolver) allocPolicyID() PolicyID {
 	return id
 }
 
-// upsertPolicy adds or updates all entries for the given policy ID in BPF maps.
+// upsertPolicyIDInBPF adds or updates all entries for the given policy ID in BPF maps.
```
Andreagit97: As a suggestion for the next time: I would do renames in a separate commit so that it is easier to understand what is really changed and not just renamed

internal/resolver/policy.go
```diff
@@ -204,46 +235,79 @@ func (r *Resolver) handleWPUpdate(wp *v1alpha1.WorkloadPolicy) error {
...
 	if !exists {
 		return fmt.Errorf("workload policy does not exist in internal state: %s", wpKey)
 	}
-	newContainers, err := r.syncWorkloadPolicy(wp)
-	if err != nil {
+
+	var newContainers policyByContainer
+	if newContainers, err = r.syncWorkloadPolicy(wp); err != nil {
+		// syncWorkloadPolicy already rolled back its BPF state internally.
+		r.setPolicyStatus(wp, agentv1.PolicyState_POLICY_STATE_ERROR, err.Error())
```
Andreagit97: Why don't we delegate the rollback and the status update to the `defer` function?

internal/resolver/policy_test.go
```diff
@@ -119,12 +152,121 @@ func TestHandleWP_Lifecycle(t *testing.T) {
...
+			workloadType: "Deployment",
+			labels:       map[string]string{v1alpha1.PolicyLabelKey: "example"},
+		},
+		containers: map[ContainerID]*containerInfo{
+			cid1: {cgID: 100, name: c1},
+		},
+	}
+	r.mu.Unlock()
+
+	handlers := r.PolicyEventHandlers()
```
Andreagit97: why don't we just call `require.Error(t, r.handleWPAdd(wp))`?

test/e2e/main_test.go
```diff
@@ -365,7 +372,18 @@ func getMainTest() types.Feature {
...
+			// Delete test-policy.
+			testPolicy := v1alpha1.WorkloadPolicy{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-policy", Namespace: workloadNamespace},
+			}
+			_ = r.Delete(ctx, &testPolicy)
+			err = wait.For(
+				conditions.New(r).ResourceDeleted(&testPolicy),
+				wait.WithTimeout(DefaultOperationTimeout),
+			)
+			require.NoError(t, err, "test-policy WorkloadPolicy should be deleted")
```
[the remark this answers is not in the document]
kyledong-suse: Actually this shouldn't happen, because in normal operation, pod deletion triggers cgroup cleanup in `RemovePodContainerFromNri`. Is this a race condition? @Andreagit97

internal/resolver/resolver.go
```diff
@@ -24,7 +24,7 @@ type Resolver struct {
 	cgroupIDToPodID map[CgroupID]PodID
 
 	nextPolicyID                PolicyID
-	wpState                     map[namespacedPolicyName]policyByContainer
+	wpState                     map[NamespacedPolicyName]wpInfo
```
Andreagit97: I would use a pointer, to be honest, so that when we get it, we can rewrite it in place instead of copying and replacing it. After this, some code paths should become easier
```suggestion
	wpState                     map[NamespacedPolicyName]*wpInfo
```

internal/resolver/policy.go
```diff
@@ -202,48 +222,79 @@ func (r *Resolver) handleWPUpdate(wp *v1alpha1.WorkloadPolicy) error {
...
+		}
+		if err := r.clearPolicyIDFromBPF(polID); err != nil {
+			return fmt.Errorf("failed to clear policy for wp %s, container %s: %w", wpKey, containerName, err)
+		}
+	}
 	return nil
 }
 
+// Must be called with the resolver lock held.
+func (r *Resolver) deleteWorkloadPolicy(wp *v1alpha1.WorkloadPolicy) error {
```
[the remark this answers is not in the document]
kyledong-suse: make sense to me.

internal/resolver/resolver.go
kyledong-suse: done

internal/resolver/policy.go
```diff
@@ -319,15 +355,30 @@ func (r *Resolver) PolicyEventHandlers() cache.ResourceEventHandler {
...
+	for k, v := range r.wpState {
+		statuses[k] = v.status
+	}
+	return statuses
+}
+
+// setPolicyStatus updates the status for the given workload policy.
+// If the policy is not in wpState, a WPInfo with empty polByContainer is stored,
+// so that ERROR state can still be reported via GetPolicyStatuses.
+func (r *Resolver) setPolicyStatus(wp *v1alpha1.WorkloadPolicy, state agentv1.PolicyState, message string) {
```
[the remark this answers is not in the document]
kyledong-suse: If we don't consider rollback, `r.wpState[wpKey]` will be always exist in the call. Since I changed `wpInfo` to pointer, so it makes more sense to simplify `setPolicyStatus` now.

internal/resolver/policy.go
```diff
@@ -202,45 +227,52 @@ func (r *Resolver) handleWPUpdate(wp *v1alpha1.WorkloadPolicy) error {
...
 	}
-	newContainers, err := r.syncWorkloadPolicy(wp)
-	if err != nil {
+
+	var err error
+	defer func() {
+		if err != nil {
+			info.setPolicyStatus(wp, agentv1.PolicyState_POLICY_STATE_ERROR, err.Error())
+		}
+		r.mu.Unlock()
```
Andreagit97: Same issue as before with the lock

test/e2e/main_test.go
```diff
@@ -141,7 +141,14 @@ func getMainTest() types.Feature {
 				err := r.Create(ctx, &policy)
 				require.NoError(t, err, "create policy")
 
-				// todo!: we should check the status of the WP is updated
+				waitForWorkloadPolicyStatusToBeUpdated(ctx, t, policy.DeepCopy())
+				var updated v1alpha1.WorkloadPolicy
```
Andreagit97: I'm not sure why we are repeating this code here; it seems to be what `waitForWorkloadPolicyStatusToBeUpdated` does. Am I missing anything?

test/e2e/main_test.go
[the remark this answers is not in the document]
kyledong-suse: It's reproduced every time. I have removed the last step of deleting policy from main_test.go. You can see the e2e test failure in this PR.

flavio (approved):

Issue #91, opened 2025-12-15
send ebpf debug logs to userspace


Andreagit97:
Today, if there is an issue at runtime on the ebpf side, there is no way to collect extra debug info. Ideally we should use a dedicated ringbuffer just to send EBPF logs to userspace and maybe propagate them as OTEL events.

Some reference
* We should send the event here, for example, https://github.com/neuvector/runtime-enforcer/blob/3f1aa14381817e539aee913d5b590a0d8610a059/bpf/main.c#L351
* We should enable this debug mode with the already present load time config https://github.com/neuvector/runtime-enforcer/blob/3f1aa14381817e539aee913d5b590a0d8610a059/bpf/load_conf.h#L7
 * We should receive this event in userspace https://github.com/neuvector/runtime-enforcer/blob/3f1aa14381817e539aee913d5b590a0d8610a059/internal/bpf/debugging.go#L3

It's not always possible to send messages to user space due to performance concern. One approach is to store error metrics in a map, so userspace can retrieve them when it's needed.

## Acceptance Criteria

- [x] Define what we want to track: which code paths do we care
- [x] Define what mechanism we want to use: counter, logs from kernel space → users space?
- [ ] [Implement this logging](https://github.com/rancher-sandbox/runtime-enforcer/issues/89#issuecomment-3852652074)
- [x] Let the agent print the `uname -a` at startup

PR #266, opened 2026-02-13
feat: configure ebpf verifier verbosity from helm chart

Andreagit97:

On some kernels, `ebpf.LogLevelBranch` and `ebpf.LogLevelBranch` verbosity levels can cause the bpf load process to hang. 
We observed it on this kernel `5.15.0-1084-aws #91~20.04.1-Ubuntu.`
For this reason, we let users choose which kind of verbosity they want to use. By default, we set the lowest verbosity `ebpf.LogLevelStats` that shouldn't cause any issue.

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/templates/agent/daemonset.yaml
+++ b/charts/runtime-enforcer/templates/agent/daemonset.yaml
@@ -34,6 +34,7 @@ spec:
         {{- end }}
         - --grpc-port={{ .Values.agent.agent.grpcExporterPort }}
         - --grpc-mtls-cert-dir={{ include "runtime-enforcer.grpc.certDir" . }}
+        - --log-level={{ .Values.agent.agent.logLevel }}
         {{- toYaml .Values.agent.agent.args | nindent 8 }}
         command:
         - /agent
--- a/charts/runtime-enforcer/tests/daemonset_test.yaml
+++ b/charts/runtime-enforcer/tests/daemonset_test.yaml
@@ -63,3 +63,13 @@ tests:
             operator: "Equal"
             value: "experimental"
             effect: "NoSchedule"
+
+  - it: "should include log level argument"
+    set:
+      agent:
+        agent:
+          logLevel: debug
+    asserts:
+      - contains:
+          path: "spec.template.spec.containers[0].args"
+          content: "--log-level=debug"
--- a/charts/runtime-enforcer/values.yaml
+++ b/charts/runtime-enforcer/values.yaml
@@ -60,6 +60,8 @@ agent:
       tag: v0.1.0
       pullPolicy: IfNotPresent
     grpcExporterPort: "50051"
+    # agent logger level. Can be one of: debug, info, warn, error.
+    logLevel: info
     resources:
       limits:
         cpu: 500m
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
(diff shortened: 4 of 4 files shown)


holyspectral (commented):
Overall LGTM, but I'd be a little hesitant to implement our own log level.

I don't know if we should expose the setting in helm chart at this stage given that the extra parameters can go through `.Values.agent.agent.args` already.  If we want it public facing, I think we could stick to existing ones like slog or others.  If it's not public facing, we could just use whatever `ebpf.LogLevel`.  WDYT?

internal/bpf/manager.go
```diff
@@ -157,20 +179,21 @@ func NewManager(logger *slog.Logger, enableLearning bool, eBPFLogLevel ebpf.LogL
...
-		},
-	}
-	if err = spec.LoadAndAssign(&objs, opts); err != nil {
-		return nil, fmt.Errorf("error loading objects: %w", err)
+	// We just load the objects here so that we can pass the maps to other components but we don't attach ebpf progs yet.
+	// The first time we use `LogLevelStats` as verbosity.
+	// If there is an issue we retry the loading with a higher verbosity.
+	objs, err := loadEbpfObjects(spec, ebpf.LogLevelStats)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to load eBPF objects with stats verbosity: %s\n", err.Error())
```
Andreagit97: I don't love to print directly on Stderr, but I believe this is the cleaner way we have to print line by line the verifier log so that we can copy and paste it. Example output
```txt
time=2026-02-16T13:29:11.306Z level=INFO msg="Probing eBPF features..." component=agent
time=2026-02-16T13:29:11.410Z level=INFO msg="Cgroupv2 supported controllers detected successfully" component=agent cgroup.controllers="[cpuset cpu io memory hugetlb pids rdma misc dmem]"
time=2026-02-16T13:29:11.410Z level=INFO msg="Load time configuration detected" component=agent component=ebpf-manager cgrp_fs_magic=cgroupv2 cgrp_v1_subsys_idx=0 debug_mode=0
failed to load eBPF objects with stats verbosity: verifier error: field EnforceCgroupPolicy: program enforce_cgroup_policy: load program: permission denied: stack depth 168 (4 line(s) omitted). Dump: load program: permission denied:
	func 'security_bprm_creds_for_exec' arg0 has btf_id 6145 type STRUCT 'linux_binprm'
	R2 unbounded memory access, make sure to bounds check any such access
	verification time 567 usec
	stack depth 168
	processed 439 insns (limit 1000000) max_states_per_insn 1 total_states 36 peak_states 36 mark_read 15
failed to load eBPF objects with branch verbosity: verifier error: field EnforceCgroupPolicy: program enforce_cgroup_policy: load program: permission denied: 787: (85) call bpf_map_lookup_elem#1: R2 unbounded memory access, make sure to bounds check any such access (376 line(s) omitted). Dump: load program: permission denied:
	0: R1=ctx() R10=fp0
	; int BPF_PROG(enforce_cgroup_policy, struct linux_binprm *bprm) { @ main.c:470
```

dottorblaster (approved):
LGTM

internal/bpf/manager.go
flavio: ok, I didn't realize it would have been that ugly. It's really hard to use that.

The majority of log collectors will just skip it, which I think is fine in this case.

Let's keep printing on the stderr for the time being.

Andreagit97:
rebased to solve conflicts

Issue #242, opened 2026-02-05
Track and display violations


flavio:
Currently violation events are emitted as OpenTelemetry traces. The traces have to be collected, stored and then viewed using external tooling. This doesn't offer the best user experience.

While we need to keep emitting violations as OTEL events (for the Security Hub), we should also provide a simpler mechanism that is better integrated into Kubernetes. Violations should be made accessible via the Kubernetes API.

Some possible ideas worth to be explored:

- Use the same plumbing used by the policy status propagation
- Explore the usage of OTEL
- Kubernetes events?

Finally, we have to decide where this information is going to be stored. Some possibilities:

- At the workload level, inside of a new status
- A dedicated CRD
- Kubernetes events

## Acceptance Criteria

- Write a RFC that tackle this problem

PR #275, opened 2026-02-17
docs: tracking and display of violations RFC

dottorblaster:

In the context of #242 we want to document the approach we intend to use in an RFC.

### DIFF
--- a/docs/rfc/0006-violation-reporting.md
+++ b/docs/rfc/0006-violation-reporting.md
@@ -1,0 +1,296 @@
+|              |                                                              |
+| :----------- | :----------------------------------------------------------- |
+| Feature Name | Kubernetes-Native Violation Reporting                        |
+| Start Date   | 2026-02-16                                                   |
+| Category     | Observability                                                |
+| RFC PR       | https://github.com/rancher-sandbox/runtime-enforcer/pull/275 |
+| State        | **ACCEPTED**                                                 |
+
+# Summary
+
+[summary]: #summary
+
+Add violation reporting paths so that policy violations are visible through
+standard Kubernetes primitives (`kubectl describe`, `kubectl get`) without
+requiring external observability infrastructure. Each agent buffers and
+deduplicates violation records in-memory; the controller scrapes violations
+from agents via the existing gRPC channel (`ScrapeViolations` RPC on port
+50051 with mTLS) during its periodic status sync, merges them into
+WorkloadPolicy status, and keeps the most recent 100 entries. Optionally,
+agents emit OTEL events (log records with `event_name` set) to a standalone
+OTEL collector Deployment over a TLS-encrypted OTLP channel.
+
+# Motivation
+
+[motivation]: #motivation
+
+Today, policy violations (unauthorized process executions detected by eBPF)
+are only emitted as OpenTelemetry log events to an external collector. Viewing
+them requires separate infrastructure, which is a poor experience for
+Kubernetes-native workflows where users expect `kubectl` to be the primary
+interface.
+
+By surfacing violations as status records:
+
+- Users see recent violation detail via `kubectl describe workloadpolicy`.
+- Users get an at-a-glance view of recent violations via the status subresource.
+- No external observability stack is required for basic violation visibility.
+
+## Examples / User Stories
+
+[examples]: #examples
+
+**As a cluster operator**, I want to run `kubectl describe workloadpolicy my-policy`
+and see recent violations with details (executable, pod, container, node) so I
+can quickly triage unauthorized process executions.
+
+**As a platform engineer**, I want `kubectl get workloadpolicy -o yaml` to show
+recent violation records so I can identify noisy policies at a glance.
+
+**As a security team member**, I want violation telemetry forwarded to our SIEM
+(via OpenTelemetry/OTLP through the existing pipeline) without disrupting current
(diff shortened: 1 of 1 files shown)


docs/rfc/0006-violation-reporting.md
```diff
@@ -0,0 +1,264 @@
...
+
+# Detailed design
+
+[design]: #detailed-design
+
+## Architecture
+
+```
+Agent (per node):
+  eBPF -> EventScraper -> OTEL Span  ──┐
```
Andreagit97: i would probably remove the OTEL Spans if we send the OTEL Events. Same in all other places where they are mentioned

docs/rfc/0006-violation-reporting.md
```diff
@@ -0,0 +1,264 @@
...
+  │                   metrics pipeline → deltatocumulative → prometheus exporter (:9090)
+  └─ traces pipeline → otlp exporter → external collector (telemetry.custom.endpoint)
+```
+
+All telemetry (traces and violation events) is sent to a standalone OTEL
+collector Deployment at `<release>-otel-collector.<namespace>.svc.cluster.local:4317`.
+The collector handles routing: violation events are counted into Prometheus
+metrics, and traces are forwarded to an external collector if configured.
+
+All violation handling happens on the agent. The operator is not involved in
```
Andreagit97: i was under the impression we wanted to use the operator to pull the violations from the agent through gRPC

Andreagit97: Side notes: 
* If we go with a push model, each agent needs RBAC to patch the resources
* If in the future we want to add a global counter of violations, we will conflict at every update.

docs/rfc/0006-violation-reporting.md
```diff
@@ -0,0 +1,264 @@
...
+**As a cluster operator**, I want to run `kubectl describe workloadpolicy my-policy`
+and see recent violations with details (executable, pod, container, node) so I
+can quickly triage unauthorized process executions.
+
+**As a platform engineer**, I want `kubectl get workloadpolicy -o yaml` to show
+recent violation records so I can identify noisy policies at a glance.
+
+**As a security team member**, I still want violations flowing to existing o11y
+setups via the existing OTEL span pipeline, unaffected by this change. I also
+want structured OTEL events for integration with log-based alerting.
```
flavio: I would rephrase that to:

```suggestion
**As a security team member**, I want violation telemetry forwarded to our SIEM (via OpenTelemetry/OTLP through the existing pipeline) without disrupting current exports, and I want violations emitted as structured OTEL events so we can drive detections and alerts.
```

docs/rfc/0006-violation-reporting.md
flavio: Same here, I think we can reuse the gRPC channel that exists between our controller and each agent to scrape the violations.

The collector would scrape the violations from all the agents and then update the WorkloadPolicy status accordingly.

In this way we would reduce concurrent access to the WorkloadPolicy resource and avoid broader RBAC permissions to the agents.

I think the idea was:

- agent keeps a list of violations, tries to deduplicate them (if an attacker attempts to start 100 times the `curl` command inside of pod XYZ, the agent returns 1 violtion instead of 100 of them)
- the controller scrapes the violations from the agent
- once scraped, the agent throws away the list of violations
- the controller adds the violations to the bottom of the list of recent violations. Removes X violations from the head of the list to ensure there are at most 100 items into the list

docs/rfc/0006-violation-reporting.md
```diff
@@ -0,0 +1,264 @@
...
+    Timestamp      metav1.Time `json:"timestamp"`
+    PodName        string      `json:"podName"`
+    ContainerName  string      `json:"containerName"`
+    ExecutablePath string      `json:"executablePath"`
+    NodeName       string      `json:"nodeName"`
+    Action         string      `json:"action"`
+}
+
+type ViolationStatus struct {
+    Violations []ViolationRecord `json:"violations,omitempty"`
```
holyspectral: When the `Violations` is full, either when it reaches `MaxViolationRecords`, or the CR reaches its maximum size, what is the expected behavior?

IMO we should purge from the oldest items, otherwise a noisy pod could completely prevent this feature from working.

holyspectral (commented):
Thanks for doing this!

Andreagit97 (commented):
LGTM!

holyspectral (commented):
LGTM!  I still have a question about the in-memory buffer but everything else looks great to me.

docs/rfc/0006-violation-reporting.md
```diff
@@ -56,48 +60,49 @@ want structured OTEL events for integration with log-based alerting.
...
+metrics.
 
 ## Agent side
 
-### ViolationReporter (`internal/agent/violation_reporter.go`)
+### ViolationBuffer (`internal/violationbuf/buffer.go`)
 
-A component that implements `manager.Runnable`. It receives violation info
-in-process from the EventScraper (no gRPC/OTLP needed) and:
+A thread-safe in-memory buffer that deduplicates repeated violations:
```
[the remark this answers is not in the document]
dottorblaster: Even loss during rolling update is fine for me as well, I'll update with a comment on that, thank you :+1:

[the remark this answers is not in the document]
flavio: fine with me as well

PR #283, opened 2026-02-17
feat(helm): allow users to configure PodDisruptionBudgets

### LINKED ISSUE #277: feat(helm): allow users to configure PodDisruptionBudgets for agent DaemonSet
Without a PDB, kubectl drain or cluster autoscaler operations can evict agent pods without restriction. This creates windows where nodes have no runtime enforcement, which is especially problematic for protect mode policies where the agent is actively blocking unauthorized executables.

We also want to be able to specify PDB for the controller.

## Proposed solution

Add optional PDB configuration to `values.yaml` (or something similar):

```yml
operator:                                                                                                             
    replicas: 1                                             
    pdb:                                                                                                                
      enabled: false                                                                                                    
      minAvailable: 1
agent:
  pdb:
    enabled: false
    # maxUnavailable is the only field that makes sense for a DaemonSet PDB.
    # Can be an integer or a percentage string (e.g. "10%").
    maxUnavailable: 1
```

And create corresponding templates.

## Notes

- PDB support for DaemonSets requires Kubernetes 1.31+ (graduated to stable via KEP-3017). Are we going to document this? Bear in mind that this would bring our minimum supported version to K8s 1.31.
- PDBs are disabled by default to avoid surprising users on clusters running older Kubernetes versions.

kyledong-suse:

Allow users to configure PodDisruptionBudgets for agent DaemonSet and operator Deployment.
This PR also fixed one bug as the `replicas` lives under `operator`, not `operator.manager`

fixes #277 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/values.yaml
+++ b/charts/runtime-enforcer/values.yaml
@@ -14,19 +14,28 @@ operator:
       repository: ghcr.io/rancher-sandbox/runtime-enforcer/operator
       tag: v0.1.0
       pullPolicy: IfNotPresent
+    # To make the Pods "Guaranteed" (evicted last under node pressure), kubelet requires
+    # requests and limits are specified for all the containers and they are equal.
+    # Please refer to https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#pod-selection-for-kubelet-eviction
     resources:
       limits:
         cpu: 500m
         memory: 128Mi
       requests:
-        cpu: 10m
-        memory: 64Mi
+        cpu: 500m
+        memory: 128Mi
     wpStatusUpdateInterval: 30s
   podSecurityContext:
     runAsNonRoot: true
     seccompProfile:
       type: RuntimeDefault
   replicas: 1
+  # # PodDisruptionBudget for the operator Deployment. Disable by default.
+  # # When enabled, limits voluntary disruption (e.g. kubectl drain).
+  # pdb:
+  #   enabled: false
+  #   # minAvailable: at least 1 pod must stay available during a disruption.
+  #   minAvailable: 1
   serviceAccount:
     annotations: {}
   tolerations: []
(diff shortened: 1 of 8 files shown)


dottorblaster (approved):
LGTM

charts/runtime-enforcer/values.yaml
```diff
@@ -27,6 +27,9 @@ operator:
     seccompProfile:
       type: RuntimeDefault
   replicas: 1
+  pdb:
```
Andreagit97: I'm a little bit worried we will block the drain node process by doing that. We have just 1 replica, and we are asking to keep it always alive. What is the expected behavior?

flavio:
Sorry folks, this is my fault... we should have done some careful planning before.

I think using PDB can work, but only for the controller.


What we should do instead.

# Define `PriorityClass`

We should define a [`PriorityClass`](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) for our agents and for our controller.

# Make our containers "guaranteed"

Take a look at how kubelet does [pod eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#pod-selection-for-kubelet-eviction).

Pods that are marked as "guaranteed" are evicted last.
A pod is "guaranteed" when:

> Guaranteed pods are guaranteed only when requests and limits are specified for all the containers and they are equal.

Which means our values.yaml should have resources that look like that:

```yaml
resources:
  requests:
    cpu: "Xm"
    memory: "YMi"
  limits:
    cpu: "Xm"      # Must match request
    memory: "YMi"  # Must match request
```

We should also document that.

# PDB

We should not do that on the DaemonSet, as others pointed out this doesn't make sense.

We should do that only on the controller.

charts/runtime-enforcer/templates/agent/priorityclass.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.priorityClass.agent.enabled }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ .Values.priorityClass.agent.name }}
+  labels:
+    {{- include "runtime-enforcer.labels" . | nindent 4 }}
+{{- $pcSpec := omit .Values.priorityClass.agent "enabled" "name" }}
+{{- $pcSpec | toYaml | nindent 0 }}
+{{- end }}
```
holyspectral: While agent could be more important than operator, do we really need two priority classes?

kyledong-suse: I think we can change it to one shared PriorityClass which is used for both operator and agent. WDYT?

PR #286, opened 2026-02-18
feat: add linter to bpf codes

### LINKED ISSUE #100: Add linter for bpf programs

holyspectral:

1. Moved to https://github.com/cpp-linter/cpp-linter-hooks for pre-commit hooks.
2. Fixed linter issues reported.
3. Upgrade clang-format and clang-tidy version to 19, to sync with what we use in building images.
4. Migrate pre-commit to cpp-linter-hooks.

Note: rev in .pre-commit-config.yaml is kept as version so we will see updatecli creates a PR to change it to hash.

fixes #100 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/updatecli/updatecli.d/pre-commit.yaml
+++ b/updatecli/updatecli.d/pre-commit.yaml
@@ -1,13 +1,12 @@
 name: Update .pre-commit-config.yaml
 sources:
-  clangformatTag:
+  cppLinterTag:
     kind: gittag
     spec:
       key: name
-      url: https://github.com/pre-commit/mirrors-clang-format
+      url: https://github.com/cpp-linter/cpp-linter-hooks
       versionFilter:
-        kind: semver
-        pattern: 18.1.*
+        kind: latest
   golangcilintTag:
     kind: gittag
     spec:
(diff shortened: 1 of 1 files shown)


updatecli/updatecli.d/pre-commit.yaml
```diff
@@ -1,28 +1,26 @@
...
         kind: latest
-  clangformatVersion:
+  cppLinterVersion:
     kind: gittag
     spec:
       key: hash
-      url: https://github.com/pre-commit/mirrors-clang-format
+      url: https://github.com/cpp-linter/cpp-linter-hooks
       versionFilter:
-        kind: semver
```
Andreagit97: What I liked about this is that we could pin the major and minor versions. It is possible that between clang-19 and clang-20, there will be some differences in how the code is formatted, but since the change shouldn't be so frequent i think we can accept it.
If we realize this something we don't want to happen too frequently we can also disable it in the future

PR #292, opened 2026-02-20
feat: implement logging ebpf side and print kernel version at agent startup

### LINKED ISSUE #91: send ebpf debug logs to userspace

Andreagit97:

Implement logging EBPF side and print kernel version at agent startup

fixes #91 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/bpf/manager.go
+++ b/internal/bpf/manager.go
@@ -135,6 +135,8 @@ func NewManager(logger *slog.Logger, enableLearning bool) (*Manager, error) {
 		return nil, fmt.Errorf("failed to remove memlock: %w", err)
 	}
 
+	logger.Info("Detected kernel version", "version", kernels.GetCurrKernelVersionStr())
+
 	logger.Info("Probing eBPF features...")
 	if err := probeEbpfFeatures(); err != nil {
 		return nil, fmt.Errorf("failure during eBPF feature probing: %w", err)
--- a/internal/kernels/kernels.go
+++ b/internal/kernels/kernels.go
@@ -14,27 +14,25 @@ import (
 var (
 	//nolint:gochecknoglobals // it makes sense to have a global variable for current kernel version
 	currKernelVersion int
+	//nolint:gochecknoglobals // it makes sense to have a global variable for current kernel version string
+	currKernelVersionStr string
 )
 
 //nolint:gochecknoinits // we need to initialize currKernelVersion at init time
 func init() {
 	// at init time, determine the current kernel version
-	var err error
-	currKernelVersion, err = getKernelVersionFromSystem()
-	if err != nil {
-		panic("unable to determine kernel version from system: " + err.Error())
-	}
-}
-
-func getKernelVersionFromSystem() (int, error) {
 	var uname unix.Utsname
 	if err := unix.Uname(&uname); err != nil {
-		return 0, err
+		panic("unable to determine kernel version from uname: " + err.Error())
 	}
-	release := strings.TrimSuffix(
-		strings.Split(unix.ByteSliceToString(uname.Release[:]), "-")[0],
-		"+")
-	return int(KernelStringToNumeric(release)), nil
+	currKernelVersionStr = unix.ByteSliceToString(uname.Release[:])
+	// we need to strip the trailing "+" from the kernel version string
+	// 6.14.0-37-generic -> 6.14.0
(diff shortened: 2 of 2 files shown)


internal/bpf/logging.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		buf := bytes.NewBuffer(record.RawSample)
+		var evt bpfLogEvt
+		if err = binary.Read(buf, binary.LittleEndian, &evt); err != nil {
+			m.logger.ErrorContext(ctx, "parsing ringbuf event", "error", err)
+			continue
+		}
+		m.logHandler(ctx, m.logger, &evt)
+	}
```
holyspectral: rate limiter

holyspectral (commented):
I think it looks great.  I was under an impression that we want to redirect `bpf_printk()`.  Redirecting only errors doesn't sound too bad.

I do still have some concern about errors returned from `bpf_ringbuf_output()` for execve events due to its potential amount.  If somehow our userspace process is slow to handle those events, it seems like we would get lots of errors in the container's log continuously.  Some ideas include a simple rate limiter in userspace, or we don't send the errors from `bpf_ringbuf_output()` to userspace.  WDYT?

internal/bpf/manager.go
```diff
@@ -130,11 +135,41 @@ func loadEbpfObjects(spec *ebpf.CollectionSpec, level ebpf.LogLevel) (*bpfObject
...
+}
+
+func GetKernelVersionFromUname() (string, error) {
+	var uname unix.Utsname
+	err := unix.Uname(&uname)
+	if err != nil {
+		return "", err
+	}
+	return unix.ByteSliceToString(uname.Release[:]), nil
+}
```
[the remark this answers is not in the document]
Andreagit97: oh i didn't know it was something Ubuntu specific. I took inspiration from tetragon https://github.com/cilium/tetragon/blob/7f09bfd2ca0123867bc7c2ab0155ac8efbc29ede/pkg/kernels/kernels.go#L27 I'm ok with using `/proc/version` and then fallback to `uname`

internal/bpf/manager.go
```diff
@@ -130,11 +134,28 @@ func loadEbpfObjects(spec *ebpf.CollectionSpec, level ebpf.LogLevel) (*bpfObject
...
 }
 
+func GetKernelVersionFromProc() (string, error) {
+	path := cgroups.DefaultProcFSPath + "/version"
+	version, err := os.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to read kernel version from '%s': %w", path, err)
+	}
+	// we need to trim the final `\n`
+	return string(version[:len(version)-1]), nil
```
kyledong-suse: I think we can take advantage of [TrimSpace](https://pkg.go.dev/strings#TrimSpace). 
Another question, if after trimming, the version is empty for some reason, should we return an error?

Andreagit97: done, thanks! :) 

> Another question, if after trimming, the version is empty for some reason, should we return an error?

I would say we can proceed. In the end, our agent should work in most cases, and an empty string would mean a bug in the /proc interface

PR #294, opened 2026-02-23
feat: implement learn configuration

### LINKED ISSUE #29: Feature Request: allow users to customize what namespaces to learn the behavior
### Is your feature request related to a problem?

Currently we monitor all namespaces and generate policy proposals based on workload on them.  Users may trust some of the components and don't want to monitor these resources. 

### Solution you'd like

Have something like a namespace selector, so users can define what namespaces to monitor.  This can potentially be implemented with neuvector/runtime-enforcement#101 

### Alternatives you've considered

_No response_

kyledong-suse:

This PR adds namespace-scoped learning via an optional NamespaceSelector on LearningConfiguration. Learning runs only in namespaces that match the selector; when the selector is unset, behavior is unchanged (learning will be applied to all namespaces).

fixes #29 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


dottorblaster (changes requested):
Next steps:

- Drop the LearningConfiguration CRD
- Allow the user to define the namespaces they want to generate proposal for in the helm chart as a labelSelector
- The namespace list flows to the agent as a CLI flag
- The agent has the responsibility to decide (as it is right now too) whether the incoming request belongs to a labeled namespace
- Do not clean up resources for namespaces for which the label is dropped :+1:

sample_apiserver_violation_exceptions.list
Andreagit97: This is a generated file that we shouldn't commit

Andreagit97: I don't recall who generates it, but maybe we can add it to `.gitignore` (?)

kyledong-suse: I think you're right, not sure if it's generated by `make generate`. I'll remove it.

internal/eventhandler/learning_controller.go
```diff
@@ -172,6 +201,29 @@ func (r *LearningReconciler) Reconcile(
...
+) (bool, error) {
+	if selector == nil {
+		return true, nil
+	}
+
+	var ns metav1.PartialObjectMetadata
+	ns.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Namespace"))
+	if err := client.Get(ctx, types.NamespacedName{Name: namespace}, &ns); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
```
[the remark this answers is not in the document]
Andreagit97: > Or we can simply skip that not-found namespace, not learning from it

this one

internal/eventhandler/learning_controller.go
```diff
@@ -172,6 +201,29 @@ func (r *LearningReconciler) Reconcile(
...
+	namespace string,
+	selector labels.Selector,
+) (bool, error) {
+	if selector == nil {
+		return true, nil
+	}
+
+	var ns metav1.PartialObjectMetadata
+	ns.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Namespace"))
+	if err := client.Get(ctx, types.NamespacedName{Name: namespace}, &ns); err != nil {
```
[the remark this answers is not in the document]
kyledong-suse: I don't think it's guaranteed that our manager will create the informer at the first Get.
It makes sense to ensure the existence for the namespace informer in `setupLearningReconciler`.

cmd/agent/main.go
```diff
@@ -87,11 +90,27 @@ func setupLearningReconciler(
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid learning-namespace-selector %q: %w", config.learningNamespaceSelector, err)
+		}
+		nsSelector = selector
+	}
+
+	// Ensure the namespace informer exists so Reconcile uses the cache instead of the API on each event.
+	if _, err := ctrlMgr.GetCache().GetInformer(ctx, &corev1.Namespace{}); err != nil {
+		return nil, fmt.Errorf("init namespace informer: %w", err)
+	}
```
holyspectral: Could you double check if we really need this?  I'm under an impression that calling `r.Client.Get()` will make controller runtime automatically track the namespace.

charts/runtime-enforcer/values.yaml
```diff
@@ -100,6 +100,8 @@ learning:
   # learning.enabled controls whether learning mode is enabled for the entire cluster.
   # When disabled, no process learning occurs. Process enforcer remains unaffected.
   enabled: true
+  # Label selector for namespaces to include in learning (empty = all).
+  namespaceSelector: ""
```
holyspectral: I was expecting something different in this field and it turned out I have to use things like `app=xxx` in this field instead of a label selector in yaml or json.  Either way, I'd recommend adding some examples for this config.  

You can also convert the whole yaml dict to a json string using helm's `toJson` built-in function, if you want, then we can accept yaml-style config here.

holyspectral (commented):
I was a little confused about the selector setting when I played with it in my environment, but overall LGTM.  Just some area that might be worth polishing.

kyledong-suse:
@flavio absolutely, I'll create a PR to update the docs today.

charts/runtime-enforcer/values.yaml
```diff
@@ -100,6 +100,18 @@ learning:
...
   # When disabled, no process learning occurs. Process enforcer remains unaffected.
   enabled: true
+  # Label selector for namespaces to include in learning (empty = all).
+  # You can set either a string or a YAML/JSON object.
+  #
+  # String format:
+  #   namespaceSelector: ""         # learn from all namespaces
+  #   namespaceSelector: "env=prod" # namespaces with label env=prod
+  #
+  # YAML/JSON object format:
```
holyspectral: Sorry I didn't have a chance to take a relook.  Just a wording that I think can be improved. 
```suggestion
  # YAML object format:
```

kyledong-suse: I can change it in the docs PR

PR #297, opened 2026-02-24
chore: fix gofmt and linter issues

holyspectral:

Just noticed that many golang linters actually do not run because we run a fast mode.  

This PR does a few things:
- Add golangci-lint-fmt into pre-commit, so it catches gofmt issues.
- Fix the existing gofmt issues.
- Moving to golangci-lint-full in pre-commit.
- Fix the linter issues (but most are false positives from https://github.com/securego/gosec/issues/1525)

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


flavio (commented):
Is the CI running with the same configuration?

PR #333, opened 2026-03-03
feat: disallow additional properties in helm

holyspectral:

- remove redundent level of values as we discussed.
- disallow additional properties via `values.schema.json`, so typo or other errors can be caught. 
- add test for service account's annotations

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Andreagit97 (approved):
Thank you for this! Just a suggestion for the first time, please try to split the renaming/removal of fields from other changes, like the JSON schema thing. It's not for the git history; it's more about simplifying reviews.

@dottorblaster I suggest you rebase on main after this, since it will very likely cause some conflicts

dottorblaster (approved):
LGTM

PR #356, opened 2026-03-06
refactor: detach the grpc client from the wp controller

Andreagit97:

The gRPC client could be used by other parts, like the kubectl plugin or a debugger tool, so it makes sense to remove it from the controller

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


dottorblaster (approved):
Love it! Looking at the architecture I just have my doubts about having to build a factory first and then the actual client through the factory, but I won't second guess something that is already in place and working like a charm 💪

flavio (approved):

PR #392, opened 2026-03-11
feat!: prevent containers from running in NRI if we fails to apply policy on it

### LINKED ISSUE #262: Design: design a better management for policy failures
The idea of this ticket is to craft a design to handle policy rollback in case of failures.

Today, when a policy is not correctly deployed on a node, we just set the status to failed, and we leave the policy in a possibly inconsistent state. Some containers are protected, others are not.

Performing an exact rollback is not easy. These are some examples:
 
1. We have a policy with a list of binaries
    ```yaml
      rulesByContainer:
        ubuntu:
          executables:
            allowed:
              - /usr/bin/ls
              - /usr/bin/cat
    ```
    If the user updates the list of binaries and we fail during the update, we need to restore the old list of binaries, but we won't have it anymore in the WP resource. So we need to store the list somewhere. The alternative is to completely delete the policy in case of failure, but it doesn't seem like a good UX (i want to update my policy, and the result is that the policy is no longer enforced...)
2. A new deployment with 3 replicas is associated with an existing policy. 2 out of 3 pods are correctly bound to the policy, the third one fails. What should we do?
3. ...

A possible idea is to turn the WP informer into a full controller so that we can retry in case of policy deployment failures. This doesn't address the case in which a new pod is added to an existing policy and the binding fails for some reason.

An initial attempt was done here https://github.com/rancher-sandbox/runtime-enforcer/pull/252

holyspectral:

As part of error handling, when we failed to apply protection on a container, we will fail the container creation flow by default.  Users can override this behavior via NRI_FAILOPEN env var.

When a container is prevented from starting, logs can be seen in these places:

In our log:
```
{"time":"2026-03-11T20:57:12.766551002Z","level":"ERROR","msg":"Runtime-enforcer has prevented the container from starting. To change this behavior, set environment variable NRI_FAILOPEN to true","component":"agent","component":"nri-handler","component":"nri-plugin","reason":"failed to add pod container from NRI","containerName":"ubuntu","podName":"ubuntu-deployment-595f9465f7-dnstl","error":"SOME ERROR"}
```

containerd:
```
Mar 11 20:59:31 kind-control-plane containerd[122398]: time="2026-03-11T20:59:31.776225421Z" level=error msg="NRI container start failed" error="rpc error: code = Unknown desc = failed to add pod container from NRI: SOME ERROR. Runtime-enforcer has prevented the container 'ubuntu-cronjob-29554377-5tzw4/ubuntu' from starting. To change this behavior, set environment variable NRI_FAILOPEN to true"
```

kubernetes:
```
    lastState:                                                          
      terminated:                                                       
        containerID: containerd://2fa1c89c84c448d36d878a44439cc2901bc2939b58be1bc5f3751be31ac4b23d
        exitCode: 128                                                   
        finishedAt: "2026-03-11T20:57:28Z"                              
        message: 'NRI container start failed: rpc error: code = Unknown desc = failed
          to add pod container from NRI: SOME ERROR. Runtime-enforcer has prevented
          the container ''ubuntu-deployment-595f9465f7-dnstl/ubuntu'' from starting.
          To change this behavior, set environment variable NRI_FAILOPEN to true'
        reason: StartError                                              
        startedAt: "1970-01-01T00:00:00Z"         
```

fixes #262

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


internal/nri/plugin.go
```diff
@@ -152,13 +158,46 @@ func (p *plugin) StartContainer(
 		"containerName", container.GetName(),
 	)
 
+	handleError := func(reason string, err error) error {
```
Andreagit97: Can we simplify this a little bit to avoid duplication?
```go
	handleError := func(reason string, err error) error {
		// fail open defaults
		var errNRI error
		msg := "container is starting WITHOUT enforcement due to NRI_FAILOPEN"
		if !p.failOpen {
			errNRI = fmt.Errorf("%s: %w. Runtime-enforcer has prevented the container '%s/%s' from starting. To change this behavior, set environment variable NRI_FAILOPEN to true",
				reason, err, pod.GetName(), container.GetName())
			msg = errNRI.Error()
		}
		p.logger.ErrorContext(
			ctx,
			msg,
			"containerID", container.GetId(),
			"containerName", container.GetName(),
			"podName", pod.GetName(),
			"podID", pod.GetUid(),
		)
		return errNRI
	}
```

holyspectral (commented):
Just noticed I didn't publish my comments...

PR #409, opened 2026-03-13
refactor: improve NRI logging

Andreagit97:

improve NRI logging

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/nri/plugin.go
+++ b/internal/nri/plugin.go
@@ -125,7 +125,7 @@ func (p *plugin) Synchronize(
 			continue
 		}
 
-		podLog := p.podLogger(pod)
+		podLogger := p.podLogger(pod)
 		containers, ok := tmpSandboxes[pod.GetId()]
 		if !ok {
 			// no containers found for pod. There are at least 2 possible reasons for this:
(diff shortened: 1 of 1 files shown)


internal/resolver/nri_api.go
```diff
@@ -88,7 +88,9 @@ func (r *Resolver) RemovePodContainerFromNri(podID PodID, containerID ContainerI
 	// remove the container from the pod
 	container, ok := state.containers[containerID]
 	if !ok {
-		r.logger.Warn("container not found", "containerID", containerID, "podID", podID)
+		// This can happen if during synchronize the container was in a not ready state and it was not sent,
+		// but then we receive the remove event for that container.
+		r.logger.Info("container not found", "containerID", containerID, "podID", podID)
```
Andreagit97: I can see this happening in our demo dev cluster, when a container is dead with OOM, it won't be sent in Synchronize

internal/nri/plugin.go
```diff
@@ -105,6 +125,7 @@ func (p *plugin) Synchronize(
 			continue
 		}
 
+		podLog := p.podLogger(pod)
```
holyspectral: I'd prefer still calling the variable `podLogger` so we know it's a slog.Logger.  Calling it just `logger` is fine with me too. 

```suggestion
		podLogger := p.podLogger(pod)
```

PR #413, opened 2026-03-16
feat(plugin): add a new command policy show protection 

### LINKED ISSUE #405: plugin: list which workloads are protected by which policy
As a goodie for the user, we want to list (using the `kubectl` plugin) which workload is protected by which policy.

## Acceptance criteria
- [ ] `kubectl runtime-enforcer list` gives a list of workloads and the associated policies
- [ ] The list can be output-ed in a table view and in JSON

Andreagit97:

This PR adds a new command `policy show protection`

fixes #405 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cmd/kubectl-plugin/policy_show_protection.go
+++ b/cmd/kubectl-plugin/policy_show_protection.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"slices"
+	"strings"
 
 	apiv1alpha1 "github.com/rancher-sandbox/runtime-enforcer/api/v1alpha1"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/podworkload"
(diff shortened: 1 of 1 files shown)


Andreagit97:
The new output is the following
```
WORKLOAD                    KIND         POLICY        MODE      STATUS
default/ubuntu-deployment   Deployment   test-policy   Monitor   Active
```

> but I'm also not sure if we can get rid of the pod view, considering some workload could have OnDelete update strategy, e.g., in DaemonSet or StatefulSet.

The logic today is a little bit simpler:
* We list all pods in the cluster.
* If the pod has a policy label, we check if it is part of the workload.
  * If it is part of the workload, we mark the workload as protected. In the end, with the VAP policy we have in the repo is not possible to protect only some pods of a workload; it's all or nothing.
  * If it is a simple pod, we just report it with KIND `Pod` and we mark it as protected

PR #447, opened 2026-03-20
feat: remove requeue, pod deletion will trigger a reconcile

kyledong-suse:

When a WorkloadPolicy is deleted while it is still in use by a Pod, a requeue log is emitted every 90 seconds, which is too frequent. This PR removes the requeue. So the behaviour will be the pod deletion will trigger a reconcile, and it'll retry finalizer removal then removes the policy.

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controller/workloadpolicy_controller.go
+++ b/internal/controller/workloadpolicy_controller.go
@@ -3,7 +3,6 @@ package controller
 import (
 	"context"
 	"fmt"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


kyledong-suse:
>If we want to open a wider discussion i believe we should disable this behavior by default; we shouldn't put a finalizer by default on the resource. IMO, it would cause a bad UX in most cases. We can let the user choose if they want to put the finalizer or not

we can discuss this topic in our daily.

internal/controller/workloadpolicy_controller.go
```diff
@@ -85,11 +80,11 @@ func (r *WorkloadPolicyReconciler) handleDeletion(
 		}
 
 		if len(podList.Items) > 0 {
-			logger.Info("Cannot remove finalizer: policy still in use by pods",
+			logger.V(1).Info("Cannot remove finalizer: policy still in use by pods",
```
Andreagit97: Uhm, this is now a debug log, I don't think this is what we wanted 🤔 
On the other side, logr doesn't allow warnings, so if I have to choose between an Info() and an Error() i would probably go with an error in this case

PR #547, opened 2026-04-15
ci: zizmor final round fixes for workflows and attestation action

dottorblaster:

Final round for Zizmor fixes :-)

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -234,4 +234,3 @@ jobs:
     permissions:
       contents: write # for updatecli to update the repository
       pull-requests: write # for updatecli to create a PR
-    secrets: inherit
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -234,4 +234,3 @@ jobs:
     permissions:
       contents: write # for updatecli to update the repository
       pull-requests: write # for updatecli to create a PR
-    secrets: inherit
```
dottorblaster, line 237: Can we get rid of this safely?

holyspectral: We probably have to follow steps defined here: https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow if the `secrets: inherit` is flagged.

holyspectral (commented):
Looks good to me!

.github/workflows/release.yml
dottorblaster: Since we're just using `GITHUB_TOKEN` in that workflow that we are invoking, it should be safe. I want to give it a try before discarding the option

holyspectral (approved):
Per discussion, let's merge this and test in the next release.

PR #553, opened 2026-04-16
feat(debugger): improve debugger logs

Andreagit97:

In a cluster with a lot of pods and nodes, printing the whole cache content for a single diff could be too much. We now just print the cache of that node.
In our demo-dev cluster, we don't install the runtime enforcer on all nodes, so we print the full cache at each iteration...
```txt
- Mismatch on number of nodes: expected 9 nodes, got 4
```

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


dottorblaster (approved):
LGTM

PR #554, opened 2026-04-16
fix: increase memory resource limit

holyspectral:

In our testing we found that 128 mb is often not sufficient when there are many nodes/workloads.

This commit increases the default number to 256 mb, so it provides better out-of-box experience.

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


dottorblaster (approved):
Thanks! 😄

Andreagit97 (approved):

PR #563, opened 2026-04-20
feat: implement oci-hook bin and injects via nri plugin

### LINKED ISSUE #483: feat: adopt an OCI hook to register containers during the learning phase
Today, container registration for the learning phase relies on the NRI StartContainer hook. While NRI fires before the container's processes begin, it operates at the containerd/CRI-O level, and the low-level runtime (runc/crun) may still start the container entrypoint before the agent has finished updating the eBPF cgroup tracker map.              

OCI runtime hooks such as CreateRuntime execute synchronously within the low-level runtime's container lifecycle, after the cgroup is created but before the container's init process starts. By injecting an OCI CreateRuntime hook via NRI's container adjustment capability, we can guarantee that the agent's eBPF maps are populated before any process runs, closing the race window and ensuring the learning phase captures every executable.

The hook binary would be a small executable deployed alongside the agent (e.g. mounted from the DaemonSet into the host). When invoked by runc/crun, it would notify the agent (e.g. via a Unix socket) to register the container's cgroup, then block until the agent confirms readiness.

### Acceptance criteria
- [ ] Implement a lightweight OCI hook executable that, when invoked as a CreateRuntime hook by the low-level runtime, the container's cgroup information to the agent over a Unix domain socket and blocks until the agent acknowledges that the cgroup has been registered in the eBPF maps
- [ ] The agent exposes a listener that receives hook notifications and registers the container's cgroup in the learning eBPF pipeline before responding
- [ ] The NRI plugin's `CreateContainer` hook injects the OCI CreateRuntime hook into the container's runtime spec via NRI container adjustment, pointing at the hook binary on the host filesystem.

kyledong-suse:

This PR adopt an OCI hook to register containers during the learning phase

fixes #483 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     strategy:
       matrix:
-        component: [controller, agent, debugger]
+        component: [controller, agent, debugger, oci-hook-installer]
         arch: [amd64, arm64]
         include:
           - arch: amd64
@@ -47,7 +47,7 @@ jobs:
       id-token: write # Signing images with cosign
     strategy:
       matrix:
-        component: [controller, agent, debugger]
+        component: [controller, agent, debugger, oci-hook-installer]
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     strategy:
       matrix:
-        component: [controller, agent, debugger]
+        component: [controller, agent, debugger, oci-hook-installer]
         arch: [amd64, arm64]
         include:
           - arch: amd64
@@ -43,7 +43,7 @@ jobs:
       id-token: write # Signing images with cosign
     strategy:
       matrix:
-        component: [controller, agent, debugger]
+        component: [controller, agent, debugger, oci-hook-installer]
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -61,7 +61,7 @@ jobs:
     name: Sign attestations and upload as artifacts
     strategy:
       matrix:
-        component: [controller, agent, debugger]
+        component: [controller, agent, debugger, oci-hook-installer]
         arch: [amd64, arm64]
         include:
           - arch: amd64
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,10 @@ kubectl-plugin-cross: ## Build kubectl plugin for all target platforms.
 debugger: generate-ebpf fmt ## Build debugger binary.
 	CGO_ENABLED=0 GOOS=linux go build -o bin/debugger ./cmd/debugger
 
+.PHONY: oci-hook
+oci-hook: fmt ## Build OCI hook binary.
+	CGO_ENABLED=0 GOOS=linux go build -o bin/oci-hook ./cmd/oci-hook
+
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/controller/main.go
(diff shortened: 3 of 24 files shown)


holyspectral (commented):
Given this is still a draft PR, just some general comments:
- I think the project structure looks great.
- You probably have to expose the unix socket in helm charts, so oci hook can access it.
- If we can support this in Tilt and e2e test that will be great.  

And IMO there are still some investigations to do:
- Let's take a look at what OCI hook can provide, especially in k8s metadata which is used for learning.
- We may need a different approach than one single unix socket, so we can support two or more agents running during rolling update.  It's unclear to me how this would change our implementation/architecture atm.  
- Let's think about how we debug/troubleshoot errors from OCI hook.

Andreagit97:
This is more or less the flow that we are using here:
- The agent starts and opens the NRI plugin
- The NRI calls our plugin with the `Synchronize` callback. The `Synchronize` callback is always called to collect the state of the node.
- The NRI calls our plugin with the `CreateContainer` callback when a new container is started.
- The plugin injects the OCI hook, giving the OCI binary path and the socket as arguments. It returns the context to the NRI. We should stay in the 2-second timeout.
- Our OCI binary is called by the container runtime on the `createRuntime` hook.
- Our OCI binary computes the cgroupID and the cgroupPath and sends them to the Agent through a Unix socket. 
- The agent populates the cgTracker EBPF map and sends a response to the OCI binary.
- The OCI binary receives the response and returns with success or failure.
- The NRI calls our plugin with the `StartContainer` callback. We already have the cgroup info computed by the OCI hook binary, but we still need to enforce the policy if necessary and update our internal caches with pod Info.
- ...

## Questions/Issues

### Synchronize callback 
This is still subject to the 2-second timeout issue, and it requires more time than all the other calls since we need to sync all the containers on the node.
  - A possible solution would be to move the work to a goroutine and return the context to the NRI. We don't need to decide if a container should start or not here, so it seems ok to work in an asynchronous way.

### OCI hook tasks

Today, the OCI binary just sends the cgroupID and the path, but the NRI plugin still needs to do a lot of work in its callback, so we are still exposed to the timeout risk.
  - Ideally, the OCI binary should send to the agent all the information that the NRI provides so that we can populate all the ebpf maps (like the policy one) inside the OCI context. The ideal solution would be to use the NRI just to `Synchronize` and to inject the OCI hook with `CreateContainer`. Does the OCI hook expose all the info we need (mainly pod labels)?
  - :warning:  IMO, this is crucial because if we still rely a lot on the NRI, using an OCI hook doesn't solve the issue; it just partially mitigates it, and so we can discuss if it is really worth it or not

### Rolling update management 

A possible solution would be to use a different Unix socket for each agent instance.

Agent1, socket1 -> NRIPlugin1 adds an OCI hook -> /opt/bin/myHook --socket socket1
Agent2, socket2 -> NRIPlugin2 adds another OCI hook -> /opt/bin/myHook --socket socket2

In this way, we have different OCI hooks that talk to different agents

holyspectral:
> Synchronize callback

I think we can already do this without OCI hook.  Maybe we could start from here? 

> Rolling update management
> A possible solution would be to use a different Unix socket for each agent instance.

My concern is more about this.  

With OCI hook, the flow becomes much more complicated and I can see conflict would happen in these areas:

- OCI hook binary
- OCI hook entries in container spec
- Endpoints for OCI hook to connect

We have to think it through to avoid missing anything.

kyledong-suse:
Thanks everyone for the discussion in our channel. Based on the discussion, we’re not going to proceed with adopting the OCI-hook approach. While it can work, it introduces significant operational and design complexity primarily to work around the NRI 2s timeout, and the rolling update behavior in particular is hard to make safe without much larger changes, which we don’t want to take on.

Instead, we’ll focus on the simpler path:
- Documenting how users can tune the NRI timeout / configure required_plugins
- Tracking the upstream NRI discussion

If we later find the timeout is a recurring production problem that can’t be addressed via configuration or upstream changes, we can revisit this with a more targeted design.

PR #568, opened 2026-04-21
chore: switch debugger base image from alpine to SUSE BCI minimal

dottorblaster:

In order to make a security audit possible I'm doing what I can on our end to ensure compliance :-) and one of the things we need is switching to BCIs where we don't use `scratch` images.

### DIFF
--- a/hack/Dockerfile.debugger.tilt
+++ b/hack/Dockerfile.debugger.tilt
@@ -1,4 +1,4 @@
-FROM registry.suse.com/bci/bci-minimal
+FROM registry.suse.com/bci/bci-base
 WORKDIR /
 COPY ./bin/debugger /debugger
 
--- a/hack/Dockerfile.debugger.tilt
+++ b/hack/Dockerfile.debugger.tilt
@@ -1,4 +1,4 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-minimal
 WORKDIR /
 COPY ./bin/debugger /debugger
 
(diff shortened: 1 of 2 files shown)


Andreagit97 (approved):

package/Dockerfile.debugger
```diff
@@ -14,8 +14,8 @@ COPY go.mod go.sum /src/
 COPY Makefile /src/
 RUN make debugger
 
-# We use alpine and not `scratch` since it provides a more complete environment.
-FROM alpine
+# We use bci-minimal and not `scratch` since it provides a more complete environment.
+FROM registry.suse.com/bci/bci-minimal
```
andreaterzolo, line 18: I suppose this image has at least a shell inside it

dottorblaster: A shell, `tcpdump`, `strace`, and all the good stuff :grimacing:

hack/Dockerfile.debugger.tilt
```diff
@@ -1,4 +1,4 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-minimal
```
holyspectral: The base image here is only used in dev deployment.  I'd use `bci-base` or keep it as `alpine`, so we can still have access to extra tools.

kyledong-suse: I agree to keep `alpine` or use `bci-base`. If we want to keep the image smaller, we can also consider with using `bci-base` as builder, and copy tools into `bci-minimal`.

kyledong-suse (approved):
LGTM, only one minor comment.

hack/Dockerfile.debugger.tilt
dottorblaster: Agreed. :saluting_face:

holyspectral (approved):
Thanks!

PR #576, opened 2026-04-23
docs: add NRI timeout and required_plugins guidance

### LINKED ISSUE #569: docs: how to adjust NRI timeout and required_plugins option
Depending on user's setup, it's possible for us to reach 2 second default NRI timeout and this would cause a few symptoms, e.g., container fails to start or agent cannot start.

To mitigate this, we should point users to change the timeout, like what [containerd](https://github.com/containerd/containerd/blob/main/docs/NRI.md) and [crio](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md#crionri-table) described, so our solution can run based on user's need. 

On the other point, some users would want to make sure that containers can't start if the required protection is not ready, e.g., due to compliance reason.  This can be done via required_plugins option in the recent container runtime.

- [containerd](https://github.com/containerd/containerd/blob/main/docs/NRI.md#default-validator)
- [crio](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md#crionridefault_validator-table)

The feature is available since containerd 2.2 (November 5, 2025) and crio 1.34.0 (August 27, 2025).

We should also document this usage.

kyledong-suse:

Document NRI timeout and required_plugins guidance 

fixes #569 

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/troubleshooting.adoc
+++ b/docs/troubleshooting.adoc
@@ -43,6 +43,80 @@ caches are aligned
 
 If there is a discrepancy you will see a diff with the affected node and pods, followed by a full dump of the agent cache for that node.
 
+== NRI timeouts and required plugins
+
+Runtime Enforcer relies on NRI (Node Resource Interface) integration provided by the container runtime.
+Some environments may hit the NRI default request timeout (default is `2s`), which can cause the NRI plugin to be detached from the container runtime. This can cause the agent to repeatedly retry connecting to the NRI plugin, resulting in the agent not coming up as expected.
+
+To mitigate this, you can increase the NRI timeout on the container runtime side by editing the container runtime configuration file.
+
+=== containerd: increase NRI timeout
+
+Edit `/etc/containerd/config.toml` and update the NRI plugin timeouts under `[plugins."io.containerd.nri.v1.nri"]`:
+
+[source,toml]
+----
+[plugins."io.containerd.nri.v1.nri"]
+  # plugin_registration_timeout is the timeout for a plugin to register after connection.
+  plugin_registration_timeout = "5s"
+  # plugin_request_timeout is the timeout for a plugin to handle an event/request.
+  plugin_request_timeout = "2s"
+----
+
+For details and additional NRI settings, see https://github.com/containerd/containerd/blob/main/docs/NRI.md#disabling-nri-support-in-containerd[containerd NRI configuration documentation].
+
+=== CRI-O: increase NRI timeout
+
+Edit `/etc/crio/crio.conf` and update the `crio.nri` table:
+
+[source,toml]
+----
+[crio.nri]
+  # Timeout for a plugin to register itself with NRI.
+  nri_plugin_registration_timeout = "5s"
+  # Timeout for a plugin to handle an NRI request.
+  nri_plugin_request_timeout = "2s"
+----
+
+For details, see https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md#crionri-table[CRI-O NRI configuration documentation].
+
+=== Required Plugins
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.adoc
```diff
@@ -43,6 +43,77 @@ caches are aligned
...
+----
+[crio.nri.default_validator]
+  nri_enable_default_validator = true
+  # List of required NRI plugins that must be present.
+  nri_validator_required_plugins = ["<list of required NRI plugins>"]
+----
+
+See https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md#crionridefault_validator-table[CRI-O NRI configuration documentation] for more details.
+
+NOTE: The *required plugins* feature is available since *containerd 2.2* (released Nov 5, 2025) and *CRI-O 1.34.0* (released Aug 27, 2025).
```
dottorblaster:
```suggestion
NOTE: The *required plugins* feature is available since *containerd 2.2* and *CRI-O 1.34.0*.
```

I would drop these release dates 👍

dottorblaster (approved):
LGTM, just a nit

docs/troubleshooting.adoc
```diff
@@ -43,6 +43,77 @@ caches are aligned
 
 If there is a discrepancy you will see a diff with the affected node and pods, followed by a full dump of the agent cache for that node.
 
+== NRI timeouts and required plugins
+
+Runtime Enforcer relies on NRI (Node Resource Interface) integration provided by the container runtime.
+Some environments may hit the NRI default request timeout (default is `2s`), which can cause the NRI plugin is detached from the container runtime. This can lead the agent to continuously retry to connect to the NRI plugin, resulting in the agent not coming up as expected.
```
andreaterzolo:
```suggestion
Some environments may hit the NRI default request timeout (default is `2s`), which can cause the NRI plugin to be detached from the container runtime. This can cause the agent to repeatedly retry connecting to the NRI plugin, resulting in the agent not coming up as expected.
```

docs/troubleshooting.adoc
```diff
@@ -43,6 +43,77 @@ caches are aligned
...
+
+To mitigate this, you can increase NRI timeouts on the container runtime side by editing the container runtime configuration file.
```
andreaterzolo:
```suggestion
To mitigate this, you can increase the NRI timeout on the container runtime side by editing the container runtime configuration file.
```

docs/troubleshooting.adoc
```diff
@@ -43,6 +43,77 @@ caches are aligned
...
+In `/etc/containerd/config.toml`, configure the `default_validator`:
+
+[source,toml]
+----
+[plugins."io.containerd.nri.v1.nri".default_validator]
+  enable = true
+  required_plugins = [ <list of required NRI plugins> ]
+----
+
+Replace `"<list of required NRI plugins>"` with the plugin name as registered with NRI in your environment.
```
andreaterzolo: Should we suggest the name of our plugin here? Not sure if we should put `00-runtime-enforcer-agent` or just  `runtime-enforcer-agent`. Moreover, we add a line saying that with this feature, we are sure that the pod cannot start without our agent running. The downside is that if the agent has some issues, no container will be started

Andreagit97 (commented):
Thank you!

docs/troubleshooting.adoc
kyledong-suse: I think we can suggest out plugin name `runtime-enforcer-agent`, and add a warning to say that with this feature enabled, pod cannot start without our agent running.

Andreagit97 (approved):

PR #581, opened 2026-04-23
feat: rework debugger code adding support for terminated containers

### LINKED ISSUE #575: [bug]: debugger pod reports cache diffs even if there are not
This is due to how NRI interaction works, as highlighted here

This is an example

```
- cache on node 'i-0ffcd91edda20f1f4' is not aligned
  []*agentv1.PodView{
  	Inverse(protocmp.Transform, protocmp.Message{
  		"@type": s"runtimeenforcer.agent.v1.PodView",
  		"containers": map[string]protocmp.Message{
  			"844da8af0ef5332ab568db9786c10fefe46ab6f26dc0df55b7a903c72d9d1b10": {"@type": s"runtimeenforcer.agent.v1.ContainerMeta", "id": string("844da8af0ef5332ab568db9786c10fefe46ab6f26dc0df55b7a903c72d9d1b10"), "name": string("wait-for-kafka")},
+ 			"d435874907dec19fbea235b42f8fdc68f456d83a91a3b3155cb4d136d152b797": s`{id:"d435874907dec19fbea235b42f8fdc68f456d83a91a3b3155cb4d136d152b797", name:"accounting"}`,
  			"e3a262b11d9d0a1f32272c0144f34efec589564d769b136270cc3d4f9f868e8c": {"@type": s"runtimeenforcer.agent.v1.ContainerMeta", "id": string("e3a262b11d9d0a1f32272c0144f34efec589564d769b136270cc3d4f9f868e8c"), "name": string("accounting")},
  		},
  		"meta": protocmp.Message{"@type": s"runtimeenforcer.agent.v1.PodMeta", "id": string("05e2a1a5-e969-4180-9d45-be34aad87ceb"), "name": string("accounting-85fd4d6c8b-8x2pt"), "namespace": string("opentelemetry-demo-chaos")},
  	}),
  	Inverse(protocmp.Transform, protocmp.Message{
  		"@type": s"runtimeenforcer.agent.v1.PodView",
  		"containers": map[string]protocmp.Message{
  			"5db958388478d49a9ffc483911c7810d13ce14c29a026c56e9dfd426ca32a8e5": {"@type": s"runtimeenforcer.agent.v1.ContainerMeta", "id": string("5db958388478d49a9ffc483911c7810d13ce14c29a026c56e9dfd426ca32a8e5"), "name": string("aws-eks-nodeagent")},
- 			"a1dbc934007866824155473c727f345b26c61ec09536aafe2268715fac1c3288": s`{id:"a1dbc934007866824155473c727f345b26c61ec09536aafe2268715fac1c3288", name:"aws-vpc-cni-init"}`,
  			"fe44cee72481408c4e1e422bb66cc0ec57e76b2e5789dbb26edfce7363d2e897": {"@type": s"runtimeenforcer.agent.v1.ContainerMeta", "id": string("fe44cee72481408c4e1e422bb66cc0ec57e76b2e5789dbb26edfce7363d2e897"), "name": string("aws-node")},
  		},
  		"meta": protocmp.Message{"@type": s"runtimeenforcer.agent.v1.PodMeta", "id": string("94bec05e-13fe-4a68-bcc4-b21bb8027185"), "name": string("aws-node-jcrv4"), "namespace": string("kube-system")},
  	}),
  	Inverse(protocmp.Transform, protocmp.Message{
  		"@type": s"runtimeenforcer.agent.v1.PodView",
  		"containers": map[string]protocmp.Message{
- 			"5cd02ac1caa315b7627619a7295344feb9a984b4d2ad47f7f5a6e783b5a78743": s`{id:"5cd02ac1caa315b7627619a7295344feb9a984b4d2ad47f7f5a6e783b5a78743", name:"flexvol-driver"}`,
  			"d276971c0b1792385d86108434e472f96ee1eefe02b19ee025e88f93e20ffba2": {"@type": s"runtimeenforcer.agent.v1.ContainerMeta", "id": string("d276971c0b1792385d86108434e472f96ee1eefe02b19ee025e88f93e20ffba2"), "name": string("calico-node")},
  		},
  		"meta": protocmp.Message{"@type": s"runtimeenforcer.agent.v1.PodMeta", "id": string("1144a2f8-5631-49b3-85fc-37bfef8ca941"), "name": string("calico-node-dmvx7"), "namespace": string("calico-system")},
  	}),
```

This is a terminated container that we still have in our agent cache because the NRI still needs to send us the `RemoveContainer` event. We don't have it in the cluster cache because the container is already terminated, and we don't cache it  (+ means we have it in our agent cache, but it is not present in the cluster)
 
```
+ 			"d435874907dec19fbea235b42f8fdc68f456d83a91a3b3155cb4d136d152b797": s`{id:"d435874907dec19fbea235b42f8fdc68f456d83a91a3b3155cb4d136d152b797", name:"accounting"}`,
```

This is an init container terminated before our agent startup, so we never received it with the Synchronize API, which is why we don't have it in our cache.

```
- 			"a1dbc934007866824155473c727f345b26c61ec09536aafe2268715fac1c3288": s`{id:"a1dbc934007866824155473c727f345b26c61ec09536aafe2268715fac1c3288", name:"aws-vpc-cni-init"}`,
```

Same for this

```
- 			"5cd02ac1caa315b7627619a7295344feb9a984b4d2ad47f7f5a6e783b5a78743": s`{id:"5cd02ac1caa315b7627619a7295344feb9a984b4d2ad47f7f5a6e783b5a78743", name:"flexvol-driver"}`,
```

The fix here is to change the way in which we compare the containers, considering also terminated/completed containers

Andreagit97:

Rework debugger code, adding support for terminated containers

fixes #575 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


internal/debugger/cluster_pod_view.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+const staticPodAnnotation = "kubernetes.io/config.mirror"
+
+type containerID = string
+
+type clusterContainerMeta struct {
+	id          containerID
+	name        string
+	terminating bool
```
kyledong-suse:
```suggestion
	terminated bool
```
It is set from `containerStatus.State.Terminated != nil`, so it's better to rename it to `terminated` all through in your PR.

internal/debugger/validate.go
```diff
@@ -0,0 +1,273 @@
...
+	clusterCaches := make(map[nodeName][]*clusterPodView)
+	var podList corev1.PodList
+	if err = c.List(ctx, &podList); err != nil {
+		return fmt.Errorf("failed to list pods in the cluster: %w", err)
+	}
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+
+		// We only consider pods that are running or succeeded
+		if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodSucceeded {
```
kyledong-suse: I wonder if we should include `corev1.PodFailed` as well in the debugger. It will likely add some noise, but since this is a debugging tool, having visibility into failed pods could be valuable. WDYT?

kyledong-suse (commented):
Overall, LGTM, just left couple of comments.

internal/debugger/validate.go
andreaterzolo: Uhm, it's not easy to decide. The goal of the debugger is to understand if the content of our agent caches is in line with what we see in the cluster. 

I can see at least a case in which a failed pod won't be present in our agent cache, but only in the cluster -> if during a `StartContainer` we prevent the container startup, we won't see the pod in our agent cache, but we will see it in the cluster as failed. This could cause a desync between the agent cache and the cluster representation.

On the other side, I can also see the opposite case, where a failed pod is present in our cache but we don't find it in our cluster representation because we skip it here...

Since the second case should be more common than the first one, I would say let's add it and see how it goes. Thank you for the suggestion!

internal/debugger/validate.go
```diff
@@ -0,0 +1,282 @@
...
+		if clusterContainer.name != cacheContainer.GetName() {
+			differences = append(
+				differences,
+				containerNameMismatch(containerID, clusterContainer.name, cacheContainer.GetName()),
+			)
+		}
+	}
+
+	for containerID, cacheContainer := range cachePod.GetContainers() {
+		if _, ok := clusterPod.containers[containerID]; ok {
```
dottorblaster: Shouldn't we use `clusterPod.getContainers()` here?

dottorblaster (approved):
No blockers so far, just a tiny nit. Thank you for moving all that code to internal, much appreciated 😍

kyledong-suse (approved):

internal/debugger/validate.go
kyledong-suse: Thank you so much!

internal/debugger/validate.go
andreaterzolo: done thanks!

PR #585, opened 2026-04-24
feat: custom debugger pod labels/annotations

holyspectral:

While checking documents about NRI, it was noticed that we missed the options to add custom labels/annotations to debugger.  This commit fixes the issue. 

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/tests/deployment_debugger_test.yaml
+++ b/charts/runtime-enforcer/tests/deployment_debugger_test.yaml
@@ -1,7 +1,12 @@
 suite: "Debugger Deployment Tests"
 templates:
   - "templates/debugger/deployment.yaml"
-
+chart:
+  version: 1.0.0
+  appVersion: 2.0.0
+release:
+  name: my-release
+  namespace: my-namespace
 tests:
   - it: "should set nodeSelector"
     set:
@@ -41,12 +46,21 @@ tests:
           path: spec.template.metadata.annotations["annotation"]
           value: annotation
 
-  - it: "should not render podLabels and podAnnotations when empty"
+  - it: "should not render custom podLabels and podAnnotations when empty"
     set:
       debugger:
         enabled: true
     asserts:
-      - notExists:
-          path: spec.template.metadata.labels["label"]
-      - notExists:
-          path: spec.template.metadata.annotations["annotation"]
+      - equal:
+          path: spec.template.metadata.labels
+          value:
+            "helm.sh/chart": "runtime-enforcer-1.0.0"
+            "app.kubernetes.io/managed-by": "Helm"
+            "app.kubernetes.io/version": "2.0.0"
+            "app.kubernetes.io/component": "debugger"
+            "app.kubernetes.io/name": "runtime-enforcer"
+            "app.kubernetes.io/instance": "my-release"
+      - equal:
+          path: spec.template.metadata.annotations
+          value:
+            "kubectl.kubernetes.io/default-container": "debugger"
--- a/charts/runtime-enforcer/tests/deployment_debugger_test.yaml
+++ b/charts/runtime-enforcer/tests/deployment_debugger_test.yaml
@@ -25,3 +25,28 @@ tests:
     asserts:
       - notExists:
           path: spec.template.spec.nodeSelector
+  - it: "should set podLabels and podAnnotations"
+    set:
+      debugger:
+        enabled: true
+        podLabels:
+          label: label
+        podAnnotations:
+          annotation: annotation
+    asserts:
+      - equal:
+          path: spec.template.metadata.labels["label"]
+          value: label
+      - equal:
+          path: spec.template.metadata.annotations["annotation"]
+          value: annotation
+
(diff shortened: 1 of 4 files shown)


charts/runtime-enforcer/values.yaml
```diff
@@ -158,6 +158,12 @@ debugger:
     requests:
       cpu: 500m
       memory: 256Mi
+  # debugger.podLabels -- Additional labels to add to the agent pods.
```
andreaterzolo:
```suggestion
  # debugger.podLabels -- Additional labels to add to the debugger pods.
```

charts/runtime-enforcer/values.yaml
```diff
@@ -158,6 +158,12 @@ debugger:
...
+  # @schema additionalProperties:true
+  podLabels: {}
+  # debugger.podAnnotations -- Additional annotations to add to the agent pods.
```
andreaterzolo:
```suggestion
  # debugger.podAnnotations -- Additional annotations to add to the debugger pods.
```

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
```diff
@@ -25,3 +25,28 @@ tests:
...
+
+  - it: "should not render podLabels and podAnnotations when empty"
+    set:
+      debugger:
+        enabled: true
+    asserts:
+      - notExists:
+          path: spec.template.metadata.labels["label"]
+      - notExists:
+          path: spec.template.metadata.annotations["annotation"]
```
andreaterzolo: Not sure if it is valid syntax, but we should assert the root path is absent not a specific label 🤔 
```suggestion
    asserts:
      - notExists:
          path: spec.template.metadata.labels
      - notExists:
          path: spec.template.metadata.annotations
```

Andreagit97 (commented):
Thank you!

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
holyspectral: Thanks!  I had to list all labels/annotations for this case because `spec.template.metadata.labels` and `spec.template.metadata.annotations` are not empty by default.  See my last commit.

holyspectral:
I will have another PR to refactor the test a bit, because common configurations like resource, labels, annotations or securityContext can be tested with the same test file.

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
```diff
@@ -25,3 +30,37 @@ tests:
...
+          annotation: annotation
+    asserts:
+      - equal:
+          path: spec.template.metadata.labels["label"]
+          value: label
+      - equal:
+          path: spec.template.metadata.annotations["annotation"]
+          value: annotation
+
+  - it: "should not render custom podLabels and podAnnotations when empty"
```
andreaterzolo, line 49: Oh, I see, of course, we have the default labels and annotation on the pod, so we cannot assert the absence of the path `spec.template.metadata.labels`. I would probably remove this test. IMO, it's enough to test the positive case: if we add labels/annotations, they are really added. A test with all these labels has more maintenance cost than real value, WDYT? In the end, what we expect to happen if we leave the field empty `{}`

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
andreaterzolo: commented here https://github.com/rancher-sandbox/runtime-enforcer/pull/585#discussion_r3148850531

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
holyspectral: Actually I added extra tests to verify these default labels and found some inconsistency issues.  For example, these labels are not present in controller/agent but debugger:

- app.kubernetes.io/managed-by
- app.kubernetes.io/version
- helm.sh/chart
- app.kubernetes.io/instance

I've added these labels and verified them in new tests: https://github.com/holyspectral/runtime-enforcer/commit/4ad0e00709c2572ee8d6481093835e4cf9ada48e#diff-6c774fad8e0ac347e9f4fe114ee4e0ebe57a631876297101179cdd64239d6597R22

We can follow up in my next PR: https://github.com/rancher-sandbox/runtime-enforcer/pull/590

andreaterzolo: uhm I see, if this is the case, I would probably rename the test from `should not render custom podLabels and podAnnotations when empty` to `should render default podLabels and podAnnotations` WDYT?

Andreagit97 (approved):

charts/runtime-enforcer/tests/deployment_debugger_test.yaml
andreaterzolo: we can do that in the follow up PR

PR #588, opened 2026-04-24
feat: adding github actions for running e2e tests

### LINKED ISSUE #432: test: run end-to-end tests over RKE2
This case is a sub-task to run e2e tests over RKE2.

## Acceptance Criteria
- [x] Tests are ran not on pull requests but on a timely fashion, potentially weekly basis
- [x] Tests are ran on releases
- [x] Tests (as in the workflow) are manually runnable

venkateshjayagopal:

Adding e2e tests which would run at 12:15 AM on Fridays, on demand and on every release.
e2e tests will run on below OS (three different jobs)

- SUSE Linux Enterprise Server 16.0
- Red Hat Enterprise Linux 10.2 Beta (Coughlan)
- Oracle Linux Server 10.1

fixes #432 

Need to add self hosted machines to this repo.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/oracle-linux-10-e2e.yml
+++ b/.github/workflows/oracle-linux-10-e2e.yml
@@ -1,0 +1,22 @@
+name: oracle-linux-10-e2e
+on:
+ release:
+   types: published
+ schedule:
+   - cron: "15 0 * * 5"
+     timezone: "America/New_York"
+ workflow_dispatch:
+jobs:
+  run-e2e-test:
+    permissions:
+      contents: read
+    name: Run e2e test on Oracle 10.1
+    runs-on: oracle
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+    - name: Run e2e test
+      run: |
+        go clean -testcache
+        E2E_USE_EXISTING_CLUSTER=true E2E_SKIP_DEPENDENCIES= go test -v -timeout 20m ./test/e2e/
--- a/.github/workflows/redhat-linux-10-e2e.yml
+++ b/.github/workflows/redhat-linux-10-e2e.yml
@@ -1,0 +1,22 @@
+name: redhat-linux-10-e2e
+on:
+ release:
+   types: published
+ schedule:
+   - cron: "15 0 * * 5"
+     timezone: "America/New_York"
+ workflow_dispatch:
+jobs:
+  run-e2e-test:
+    permissions:
+      contents: read
+    name: Run e2e test on RHEL 10.2
+    runs-on: redhat
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+    - name: Run e2e test
+      run: |
+        go clean -testcache
+        E2E_USE_EXISTING_CLUSTER=true E2E_SKIP_DEPENDENCIES= go test -v -timeout 20m ./test/e2e/
--- a/.github/workflows/suse-linux-16-e2e.yml
+++ b/.github/workflows/suse-linux-16-e2e.yml
@@ -1,0 +1,22 @@
+name: suse-linux-16-e2e
+on:
+ release:
+   types: published
+ schedule:
+   - cron: "15 0 * * 5"
+     timezone: "America/New_York"
+ workflow_dispatch:
+jobs:
+  run-e2e-test:
+    permissions:
+      contents: read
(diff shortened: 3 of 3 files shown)


.github/workflows/oracle-linux-10-e2e.yml
```diff
@@ -0,0 +1,22 @@
+name: oracle-linux-10-e2e
+on:
+ release:
+   types: published
+ schedule:
+   - cron: "15 0 * * 5"
```
andreaterzolo, line 6: I would probably move this to Monday morning, so that we have all the week to tackle the issue. We can also consider using the European timezone so that we have more time to address issues. WDYT?
```suggestion
   - cron: "15 8 * * 1"
     timezone: "Europe/Rome"
```

.github/workflows/oracle-linux-10-e2e.yml
```diff
@@ -0,0 +1,22 @@
...
+     timezone: "America/New_York"
+ workflow_dispatch:
+jobs:
+  run-e2e-test:
+    permissions:
+      contents: read
+    name: Run e2e test on Oracle 10.1
+    runs-on: oracle
```
andreaterzolo, line 14: These 3 files seem identical apart from the runner. WDYT about  using a matrix?
```yaml
name: self-hosted-e2e
on:
  release:
    types: published
  schedule:
    - cron: "15 8 * * 1"
      timezone: "Europe/Rome"
  workflow_dispatch:
    inputs:
      target:
        description: "Target runner"
        required: true
        default: all
        type: choice
        options:
          - all
          - oracle
          - redhat
          - suse
jobs:
  run-e2e-test:
    permissions:
      contents: read
    strategy:
      fail-fast: false
      matrix:
        include:
          - label: Oracle Linux 10.1
            runner: oracle
          - label: RHEL 10.2
            runner: redhat
          - label: SLES 16
            runner: suse
    if: ${{ github.event_name != 'workflow_dispatch' || inputs.target == 'all' || inputs.target == matrix.runner }}
    name: Run e2e test on ${{ matrix.label }}
    runs-on: ${{ matrix.runner }}
    steps:
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          persist-credentials: false
      - name: Run e2e test
        run: |
          go clean -testcache
          E2E_USE_EXISTING_CLUSTER=true E2E_SKIP_DEPENDENCIES= go test -v -timeout 20m ./test/e2e/
```

.github/workflows/oracle-linux-10-e2e.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
```
andreaterzolo, line 18: I'm wondering if we should add a step to set up the Go version, otherwise we should always rely on the underlying runner to have the updated Go version. WDYT?
```yaml
      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
        with:
          go-version-file: "go.mod"
```

Andreagit97 (commented):
Thank you!

.github/workflows/oracle-linux-10-e2e.yml
dottorblaster: I agree with this actually

dottorblaster (commented):
My main remark here as @Andreagit97 already commented is about using a matrix. Thank you for this!

.github/workflows/oracle-linux-10-e2e.yml
venkateshjayagopal: Yes will fix this.

.github/workflows/oracle-linux-10-e2e.yml
venkateshjayagopal: Yes I was thinking if we can include Sam's and Kyle's fixes before we run the tests on Friday early morning. Then analyze the results on Friday.
But yes since we run on Monday it should include everyone's fixes.

.github/workflows/oracle-linux-10-e2e.yml
venkateshjayagopal: This is self hosted runner, we have to maintain the runner with latest RKE2 and go version. But this is a good idea.
I was thinking this as workaround until CAB team approves our use case to use AWS. 
In that case, we will deploy the latest go and stable RKE2 versions before we execute e2e tests.

venkateshjayagopal:
Will be using my private repo to run the e2e tests on different OS as discussed in the meeting.

PR #590, opened 2026-04-27
refactor: helm unit-test

holyspectral:

This PR refactors the current helm unit-tests to ensure common configurations can be applied to all runtime enforcement pods correctly:
   - image path/tag
   - imagePullPolicy
   - tolerations
   - env var
   - extra labels
   - extra annotations
   - nodeSelector
   - affinity

In addition, these issues are fixed along the way:

1. Implement envvars, toleration, affinity and imagePullSecrets support for debugger.
2. Add missing labels to controller/agent pods

fixes #

It's easier to review commit by commit. 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/charts/runtime-enforcer/templates/debugger/deployment.yaml
+++ b/charts/runtime-enforcer/templates/debugger/deployment.yaml
@@ -27,6 +27,7 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ include "runtime-enforcer.fullname" . }}-debugger
+      securityContext: {{- toYaml .Values.debugger.podSecurityContext | nindent 8 }}
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
       {{- toYaml .Values.imagePullSecrets | nindent 8 }}
--- a/charts/runtime-enforcer/templates/debugger/deployment.yaml
+++ b/charts/runtime-enforcer/templates/debugger/deployment.yaml
@@ -27,6 +27,10 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ include "runtime-enforcer.fullname" . }}-debugger
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml .Values.imagePullSecrets | nindent 8 }}
+      {{- end }}
       containers:
       - name: debugger
         image: {{ .Values.debugger.image.repository }}:{{ .Values.debugger.image.tag }}
@@ -40,6 +44,9 @@ spec:
           value: {{ include "runtime-enforcer.agent.labelSelectorString" . | quote }}
         - name: DEBUGGER_CERT_DIR
           value: {{ include "runtime-enforcer.grpc.certDir" . | quote }}
+        {{- with .Values.debugger.env }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
         {{- with .Values.debugger.resources }}
         resources:
         {{- toYaml . | nindent 10 }}
@@ -49,9 +56,17 @@ spec:
           mountPath: {{ include "runtime-enforcer.grpc.certDir" . }}
           readOnly: true
         securityContext: {{- toYaml .Values.debugger.containerSecurityContext | nindent 10 }}
+      {{- with .Values.debugger.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.debugger.nodeSelector }}
       nodeSelector:
       {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.debugger.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
       - name: grpc-certs
--- a/charts/runtime-enforcer/tests/daemonset_test.yaml
+++ b/charts/runtime-enforcer/tests/daemonset_test.yaml
@@ -38,8 +38,7 @@ tests:
     asserts:
       - contains:
           path: "spec.template.spec.containers[0].args"
-          content: "--learning-namespace-selector={\"matchExpressions\":[{\"key\":\"kuber\
-            netes.io/metadata.name\",\"operator\":\"Exists\"}]}"
+          content: '--learning-namespace-selector={"matchExpressions":[{"key":"kubernetes.io/metadata.name","operator":"Exists"}]}'
 
   - it: "should render a custom learning namespace selector"
     set:
--- a/charts/runtime-enforcer/tests/daemonset_test.yaml
+++ b/charts/runtime-enforcer/tests/daemonset_test.yaml
@@ -3,20 +3,17 @@ templates:
   - "templates/agent/daemonset.yaml"
 
 tests:
-  - it: "should allow image, and imagePullPolicy to be overridden"
-    set:
-      agent:
-        image:
-          repository: somewhere/agent
-          tag: v9.9.9
-          pullPolicy: Always
+  - it: "should render component and default container"
     asserts:
       - equal:
-          path: "spec.template.spec.containers[0].image"
-          value: "somewhere/agent:v9.9.9"
+          path: spec.template.metadata.labels["app.kubernetes.io/component"]
+          value: "agent"
       - equal:
-          path: "spec.template.spec.containers[0].imagePullPolicy"
-          value: "Always"
+          path: spec.template.metadata.annotations["kubectl.kubernetes.io/default-container"]
+          value: "agent"
+
+  - it: "should render default priority class name"
+    asserts:
       - equal:
           path: "spec.template.spec.priorityClassName"
           value: "runtime-enforcer-priorityclass"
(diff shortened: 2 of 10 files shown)


holyspectral:
One thing I noticed is that we don't have `nameOverride` and `fullnameOverride` defined in our values.yaml, like what we have in [_helpers.tpl](https://github.com/rancher-sandbox/runtime-enforcer/blob/65dc54ff5962f657005b2ae56d590af2657f5944/charts/runtime-enforcer/templates/_helpers.tpl#L14), therefore they're not allowed.  We are probably not interested in this at this stage, so maybe we can do that later when there is a request.

holyspectral:
This is now ready for review.

charts/runtime-enforcer/tests/deployment_test.yaml
```diff
@@ -3,24 +3,29 @@ templates:
   - "templates/controller/deployment.yaml"
 
 tests:
-  - it: "should allow image, and imagePullPolicy to be overridden"
+  - it: "should render component and default container"
+    set:
+      debugger:
+        enabled: true
```
kyledong-suse: Do we want to set the `debugger.enable: true` as we're testing the controller's deployment?

charts/runtime-enforcer/tests/daemonset_test.yaml
```diff
@@ -30,7 +27,8 @@ tests:
     asserts:
       - notContains:
           path: "spec.template.spec.containers[0].args"
-          content: '--learning-namespace-selector={"matchExpressions":[{"key":"kubernetes.io/metadata.name","operator":"Exists"}]}'
+          content: "--learning-namespace-selector={\"matchExpressions\":[{\"key\":\"kuber\
```
andreaterzolo, line 30: to be honest i would prefer to use `'` so that we don't need to escape each `"`, same for other places, WDYT? If i understand correctly you did that to limit the line length. Are you using a specific linter to do that?

charts/runtime-enforcer/templates/debugger/deployment.yaml
```diff
@@ -27,6 +27,11 @@ spec:
...
     spec:
       serviceAccountName: {{ include "runtime-enforcer.fullname" . }}-debugger
+      securityContext: {{- toYaml .Values.debugger.podSecurityContext | nindent 8 }}
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets:
+      {{- toYaml .Values.imagePullSecrets | nindent 8 }}
+      {{- end }}
       containers:
       - name: debugger
         image: {{ .Values.debugger.image.repository }}:{{ .Values.debugger.image.tag }}
```
kyledong-suse, line 37: I'd suggest we add the default to make it consistent with agent and controller.

```suggestion
        image: {{ .Values.debugger.image.repository }}:{{ .Values.debugger.image.tag | default .Chart.AppVersion }}
```

kyledong-suse (approved):
LGTM, thanks! Just couple of minor suggestions.

Andreagit97 (commented):
Thank you for this! Just a minor comment

charts/runtime-enforcer/tests/daemonset_test.yaml
holyspectral: hmm it seems like [this extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) did this.  It's a dependency of Swagger so I didn't notice it.  

I don't have strong opinion about single quote or double quote, but it does feel like we should make them consistent, probably a coding convention thing.

@Andreagit97 @kyledong-suse @dottorblaster any thoughts on this?

charts/runtime-enforcer/templates/debugger/deployment.yaml
holyspectral: Ah I didn't notice that we have this default version specified.  I do agree that we should keep them consistent, but I think we should do the opposite by removing the default specified here.  Some reasons:
1. We already have otel collector which would run a different version than `AppVersion`.
2. We probably won't have use case where `image.tag` is empty.  Even we have, the `image.repository` will probably be missing too.
3. For error handling, instead of falling back silently, I'd prefer it to fail immediately.  

WDYT?

charts/runtime-enforcer/tests/deployment_test.yaml
holyspectral: Good eyes!

dottorblaster: As Kyle already asked, do we need this?

dottorblaster (approved):
Just one comment but in general very nice work

charts/runtime-enforcer/tests/daemonset_test.yaml
dottorblaster: I think this is a good case that justifies switching to single quotes, the escaping inside is a lot and for sure it makes the string more readable :sweat_smile: 

Generally speaking I agree we should try to stick with double quotes, but in this case it has more drawbacks than actual benefits.

charts/runtime-enforcer/templates/debugger/deployment.yaml
kyledong-suse: I agree we should keep them consistent, and I’d go with removing `default .Chart.AppVersion` from agent/controller, so all three components behave the same and we don’t silently fall back.
Additionally, I wonder if we can add a schema requirement for `image.repository` and `image.tag` for a better fail fast mechanism for all of them?

charts/runtime-enforcer/tests/daemonset_test.yaml
kyledong-suse: +1 to switching to single quotes for these helm-unittest assertions. It avoids a lot of escaping and line wrapping, and is much more readable.
I agree that we should make our code consistent. Maybe we can keep double quotes as default, and use single quotes as an exception when the string contains many " characters?

holyspectral: Thanks for all the comments!  I'd say this is probably a tooling issue.  [The extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) is probably the most popular vscode plugin for yaml, but it only supports all single quotes or all double quotes.  Based on this, we probably have these options:

1. Use single quote everywhere and have document to mention this since this is not the default setting of the extension.
2. Keep using double quote everywhere and accept those extra escape characters.
3. Do nothing.  It's up to each developer to decide which they want to use.  

WDYT?

holyspectral:
Thanks everyone!  I've implemented those review comments and reverted the line mentioned [here](https://github.com/rancher-sandbox/runtime-enforcer/pull/590#discussion_r3161857289), which we can revisit once we have conclusion about the coding convention.

charts/runtime-enforcer/tests/daemonset_test.yaml
andreaterzolo: I would say let's use the formatter and find the right configuration. For example if i put the following in my settings.json
```json
	"yaml.completion": true,
	"yaml.format.enable": true,
	"yaml.format.singleQuote": true,
	"[yaml]": {
		"editor.defaultFormatter": "redhat.vscode-yaml",
	},
```

I see that the single quotes are not touched

```diff
diff --git a/charts/runtime-enforcer/tests/daemonset_test.yaml b/charts/runtime-enforcer/tests/daemonset_test.yaml
index b7362cf..d2b3c4c 100644
--- a/charts/runtime-enforcer/tests/daemonset_test.yaml
+++ b/charts/runtime-enforcer/tests/daemonset_test.yaml
@@ -177,7 +177,8 @@ tests:
       - notExists:
           path: spec.template.spec.affinity
 
-  - it: "should set OTEL env vars to built-in collector when collectorStrategy is default"
+  - it: "should set OTEL env vars to built-in collector when collectorStrategy is
+      default"
     set:
       telemetry:
         collectorStrategy: default
@@ -186,14 +187,16 @@ tests:
           path: "spec.template.spec.containers[0].env"
           content:
             name: OTEL_EXPORTER_OTLP_ENDPOINT
-            value: "https://RELEASE-NAME-runtime-enforcer-otel-collector.NAMESPACE.svc.cluster.local:4317"
+            value: "https://RELEASE-NAME-runtime-enforcer-otel-collector.NAMESPACE.svc.clus\
+              ter.local:4317"
       - contains:
           path: "spec.template.spec.containers[0].env"
           content:
             name: OTEL_EXPORTER_OTLP_CERTIFICATE
             value: "/etc/runtime-enforcer/certs/ca.crt"
 
-  - it: "should set OTEL env vars to external endpoint with insecure when no CA cert"
+  - it: "should set OTEL env vars to external endpoint with insecure when no CA
+      cert"
     set:
       telemetry:
         collectorStrategy: external
```

holyspectral: hmm I tried in my environment and it works as you said, but tbh it seems like a bug because when `"yaml.format.singleQuote": true` is there, the quote won't be updated anymore regardless it's a single quote or double quote like their document said.  I have VSCode 1.115.0 + redhat.vscode-yaml 1.22.0.

Per our discussion, I will follow up outside of this PR.

PR #600, opened 2026-04-30
build(deps): bump kubernetes dependencies to 0.36.0

dottorblaster:
Just fixing our dependency bumps :-)

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


config/webhook/manifests.yaml
holyspectral, line 1: It seems like we don't need this file at all because the charts/runtime-enforcer/templates/controller/webhook.yaml is highly patched.  Maybe consider adding config/ to .gitignore, or add `output:webhook:none` to [this line](https://github.com/rancher-sandbox/runtime-enforcer/blob/7f5ad1a92a37c0e5e1d30a7cfa0e765d9e3fd341/Makefile#L37)?

Makefile
```diff
@@ -178,7 +178,7 @@ HELM_VALUES_SCHEMA_JSON ?= $(LOCALBIN)/helm-values-schema-json
 ## Tool Versions
 CONTROLLER_TOOLS_VERSION ?= v0.17.1
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
-ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
+ENVTEST_VERSION ?= release-0.23
```
holyspectral, line 181: Do we need this?  It seems like this version would conflict with what we have in go.mod easily.

holyspectral (commented):
Thanks!  Just two comments.

Makefile
dottorblaster: For now we do, as soon as they release a new 0.24 binary I'll restore the variable :+1:

config/webhook/manifests.yaml
dottorblaster: I'll go with `output:webhook:none`, thanks :+1:

holyspectral (approved):
LGTM!  And I think the new envtest is also available now.

Andreagit97 (approved):

PR #605, opened 2026-05-05
feat: modify startup probe with INFO and final sync failure error

### LINKED ISSUE #582: Consistent error logs on startup
Every time I start a minikube cluster, I see some error logs in the agent. This is due to our startup probe

```json
{"time":"2026-04-23T14:26:07.934065725Z","level":"ERROR","msg":"WorkloadPolicy handler is not synced","component":"agent","error":"failed to list WorkloadPolicies during HasSynced check: Timeout: failed waiting for *v1alpha1.WorkloadPolicy Informer to sync"}
{"time":"2026-04-23T14:26:07.934210509Z","level":"WARN","msg":"NRI handler has not yet synchronized","component":"agent","component":"resolver"}
```

Full logs

```json
{"time":"2026-04-23T14:25:51.708053913Z","level":"INFO","msg":"OTLP telemetry enabled","component":"agent","endpoint":"https://runtime-enforcer-otel-collector.runtime-enforcer.svc.cluster.local:4317"}
{"time":"2026-04-23T14:25:51.712754721Z","level":"INFO","msg":"Detected kernel version","component":"agent","component":"ebpf-manager","version":"6.6.95"}
{"time":"2026-04-23T14:25:51.712806929Z","level":"INFO","msg":"Probing eBPF features","component":"agent","component":"ebpf-manager"}
{"time":"2026-04-23T14:25:51.89644626Z","level":"INFO","msg":"cgroup info detected","component":"agent","fs_magic":"cgroupv2","v1_subsys_idx":0,"resolution_path":"/proc/1/root/sys/fs/cgroup"}
{"time":"2026-04-23T14:25:51.896471818Z","level":"INFO","msg":"bpf load config","component":"agent","fs_magic_id":1667723888,"v1_subsys_idx":0,"debug_mode":0,"learning_enabled":1}
{"time":"2026-04-23T14:25:51.920991118Z","level":"INFO","msg":"eBPF prog and maps loaded successfully","component":"agent"}
{"time":"2026-04-23T14:26:07.92645177Z","level":"INFO","msg":"learning mode is enabled","component":"agent","namespaceSelector":"{\"matchExpressions\":[{\"key\":\"kubernetes.io/metadata.name\",\"operator\":\"Exists\"}]}"}
{"time":"2026-04-23T14:26:07.926502185Z","level":"INFO","msg":"connecting to NRI socket","component":"agent","component":"nri-handler"}
{"time":"2026-04-23T14:26:07.926969369Z","level":"INFO","msg":"starting manager","component":"agent"}
{"time":"2026-04-23T14:26:07.92702775Z","level":"INFO","msg":"Starting metrics server","component":"agent","logger":"controller-runtime/metrics"}
{"time":"2026-04-23T14:26:07.927060491Z","level":"INFO","msg":"starting server","component":"agent","name":"health probe","addr":"[::]:8081"}
{"time":"2026-04-23T14:26:07.927095738Z","level":"INFO","msg":"Serving metrics server","component":"agent","logger":"controller-runtime/metrics","bindAddress":":8080","secure":false}
{"time":"2026-04-23T14:26:07.927139211Z","level":"INFO","msg":"Starting BPF Manager...","component":"agent","component":"ebpf-manager"}
{"time":"2026-04-23T14:26:07.927529398Z","level":"WARN","msg":"NRI handler has not yet synchronized","component":"agent","component":"resolver"}
{"time":"2026-04-23T14:26:07.930346288Z","level":"INFO","msg":"Starting EventSource","component":"agent","controller":"workloadpolicy","controllerGroup":"security.rancher.io","controllerKind":"WorkloadPolicy","source":"kind source: *v1alpha1.WorkloadPolicy"}
{"time":"2026-04-23T14:26:07.93041068Z","level":"INFO","msg":"Starting EventSource","component":"agent","controller":"learningEvent","source":"channel source: 0x32df204df500"}
{"time":"2026-04-23T14:26:07.930426139Z","level":"INFO","msg":"Starting Controller","component":"agent","controller":"learningEvent"}
{"time":"2026-04-23T14:26:07.930437761Z","level":"INFO","msg":"Starting workers","component":"agent","controller":"learningEvent","worker count":1}
{"time":"2026-04-23T14:26:07.930504738Z","level":"INFO","msg":"Created plugin 00-runtime-enforcer-agent (agent, handles StartContainer,RemoveContainer)","component":"agent","component":"nri-handler","component":"nri-stub"}
{"time":"2026-04-23T14:26:07.930567927Z","level":"INFO","msg":"Registering plugin 00-runtime-enforcer-agent...","component":"agent","component":"nri-handler","component":"nri-stub"}
{"time":"2026-04-23T14:26:07.930931395Z","level":"INFO","msg":"Starting gRPC exporter","component":"agent","component":"grpc_exporter","addr":":50051","mTLS":true}
{"time":"2026-04-23T14:26:07.932343527Z","level":"INFO","msg":"Configuring plugin 00-runtime-enforcer-agent for runtime cri-o/1.35.0...","component":"agent","component":"nri-handler","component":"nri-stub"}
{"time":"2026-04-23T14:26:07.93248262Z","level":"INFO","msg":"Started plugin 00-runtime-enforcer-agent...","component":"agent","component":"nri-handler","component":"nri-stub"}
{"time":"2026-04-23T14:26:07.934065725Z","level":"ERROR","msg":"WorkloadPolicy handler is not synced","component":"agent","error":"failed to list WorkloadPolicies during HasSynced check: Timeout: failed waiting for *v1alpha1.WorkloadPolicy Informer to sync"}
{"time":"2026-04-23T14:26:07.934210509Z","level":"WARN","msg":"NRI handler has not yet synchronized","component":"agent","component":"resolver"}
{"time":"2026-04-23T14:26:07.934229004Z","level":"INFO","msg":"healthz check failed","component":"agent","logger":"controller-runtime/healthz","statuses":[{},{}]}
{"time":"2026-04-23T14:26:07.934894273Z","level":"ERROR","msg":"WorkloadPolicy handler is not synced","component":"agent","error":"failed to list WorkloadPolicies during HasSynced check: Timeout: failed waiting for *v1alpha1.WorkloadPolicy Informer to sync"}
{"time":"2026-04-23T14:26:07.934920682Z","level":"INFO","msg":"healthz check failed","component":"agent","logger":"controller-runtime/healthz","statuses":[{},{}]}
{"time":"2026-04-23T14:26:07.935845592Z","level":"INFO","msg":"Synchronizing pod sandboxes","component":"agent","component":"nri-handler","component":"nri-plugin","podCount":14,"containerCount":16}
{"time":"2026-04-23T14:26:07.937183743Z","level":"INFO","msg":"Pod sandboxes synchronized","component":"agent","component":"nri-handler","component":"nri-plugin"}
{"time":"2026-04-23T14:26:08.035448144Z","level":"INFO","msg":"Starting Controller","component":"agent","controller":"workloadpolicy","controllerGroup":"security.rancher.io","controllerKind":"WorkloadPolicy"}
{"time":"2026-04-23T14:26:08.035494512Z","level":"INFO","msg":"Starting workers","component":"agent","controller":"workloadpolicy","controllerGroup":"security.rancher.io","controllerKind":"WorkloadPolicy","worker count":1}
```

We should try to avoid printing errors at each startup

Repro

```bash
minikube start --driver=kvm2 --container-runtime=containerd
helm install runtime-enforcer ./charts/runtime-enforcer \                                                         
  --namespace runtime-enforcer \                              
  --set agent.image.tag=latest \
  --set controller.image.tag=latest \
  --create-namespace \
  --wait
```

I see 2 solutions here:
* convert the log to a warning and find a way to notify the user with an error log only if the final startup probe attempt fails
* Revisit the way in which we handle the startup probe to avoid this kind of error.

kyledong-suse:

modify startup probe with INFO and final sync failure error

fixes #582 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -39,6 +39,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+const wpSyncInProgressMsg = "waiting for WorkloadPolicy synchronization to complete"
+
 type Config struct {
 	learningNamespaceSelector string
 	nriSocketPath             string
@@ -95,25 +97,30 @@ func setupWorkloadPolicyHandler(
 	ctrlMgr manager.Manager,
 	logger *slog.Logger,
 	resolver *resolver.Resolver,
-) error {
+) (*workloadpolicyhandler.WorkloadPolicyHandler, error) {
 	wpHandler := workloadpolicyhandler.NewWorkloadPolicyHandler(ctrlMgr.GetClient(), logger, resolver)
 	err := wpHandler.SetupWithManager(ctrlMgr)
 	if err != nil {
-		return fmt.Errorf("unable to set up WorkloadPolicy handler: %w", err)
+		return nil, fmt.Errorf("unable to set up WorkloadPolicy handler: %w", err)
 	}
 	// controller-runtime doesn't support a separate startup probe, so we use the readiness probe instead.
 	// See https://github.com/kubernetes-sigs/controller-runtime/issues/2644 for more details.
 	if err = ctrlMgr.AddReadyzCheck("policy readyz", func(req *http.Request) error {
 		if syncErr := wpHandler.HasSynced(req.Context()); syncErr != nil {
-			logger.ErrorContext(req.Context(), "WorkloadPolicy handler is not synced", "error", syncErr)
-			return fmt.Errorf("WorkloadPolicy handler is not synced: %w", syncErr)
+			logger.InfoContext(
+				req.Context(),
+				wpSyncInProgressMsg,
+				"error",
+				syncErr,
+			)
+			return fmt.Errorf("%s: %w", wpSyncInProgressMsg, syncErr)
 		}
 		return nil
 	}); err != nil {
-		return fmt.Errorf("failed to add policy readiness probe: %w", err)
+		return nil, fmt.Errorf("failed to add policy readiness probe: %w", err)
 	}
 
-	return nil
+	return wpHandler, nil
 }
 
 func waitForMutatingAdmissionWebhook(ctx context.Context) error {
@@ -239,7 +246,8 @@ func startAgent(ctx context.Context, logger *slog.Logger, config Config) error {
 		return fmt.Errorf("failed to create resolver: %w", err)
 	}
 
-	if err = setupWorkloadPolicyHandler(ctrlMgr, logger, resolver); err != nil {
+	wpHandler, err := setupWorkloadPolicyHandler(ctrlMgr, logger, resolver)
+	if err != nil {
 		return err
 	}
 
@@ -298,6 +306,12 @@ func startAgent(ctx context.Context, logger *slog.Logger, config Config) error {
 
 	logger.InfoContext(ctx, "starting manager")
 	if err = ctrlMgr.Start(ctx); err != nil {
+		if !resolver.IsNRISynchronized() || !wpHandler.IsSynchronized() {
+			logger.ErrorContext(ctx, "agent terminated before startup synchronization completed",
+				"nriSynchronized", resolver.IsNRISynchronized(),
+				"workloadPoliciesSynchronized", wpHandler.IsSynchronized(),
+			)
+		}
 		return fmt.Errorf("failed to start manager: %w", err)
 	}
 
--- a/internal/nri/plugin.go
+++ b/internal/nri/plugin.go
@@ -14,6 +14,8 @@ import (
 	"github.com/rancher-sandbox/runtime-enforcer/internal/types/workloadkind"
 )
 
+const nriSyncRetryMsg = "NRI pod/container sync not ready yet, will retry"
+
 type plugin struct {
 	stub            stub.Stub
 	logger          *slog.Logger
@@ -163,8 +165,8 @@ func (p *plugin) Synchronize(
 		)
 		if err := p.resolver.AddPodContainerFromNri(podData); err != nil {
 			// This could be recoverable. Returning an error so we can retry.
-			podLogger.ErrorContext(ctx, "failed to add pod container from NRI", "error", err)
-			return nil, fmt.Errorf("failed to add pod container from NRI: %w", err)
+			podLogger.InfoContext(ctx, nriSyncRetryMsg, "error", err)
+			return nil, fmt.Errorf("%s: %w", nriSyncRetryMsg, err)
 		}
 	}
 	// Mark resolver as synchronized, so old agent can be safely removed.
--- a/internal/resolver/nri_api.go
+++ b/internal/resolver/nri_api.go
@@ -8,6 +8,8 @@ import (
 	"github.com/rancher-sandbox/runtime-enforcer/internal/bpf"
 )
 
+const nriSyncInProgressMsg = "waiting for NRI synchronization to complete"
+
 func convertPodData(pod PodInput) *podEntry {
 	return &podEntry{
 		meta:       &pod.Meta,
@@ -113,10 +115,14 @@ func (r *Resolver) NRISynchronized() {
 	r.nriSynchronized.Store(true)
 }
 
+func (r *Resolver) IsNRISynchronized() bool {
+	return r.nriSynchronized.Load()
+}
+
 func (r *Resolver) Ping(req *http.Request) error {
 	if !r.nriSynchronized.Load() {
-		r.logger.WarnContext(req.Context(), "NRI handler has not yet synchronized")
-		return errors.New("NRI handler has not yet synchronized")
+		r.logger.InfoContext(req.Context(), nriSyncInProgressMsg)
+		return errors.New(nriSyncInProgressMsg)
 	}
 	return nil
 }
--- a/charts/runtime-enforcer/templates/agent/daemonset.yaml
+++ b/charts/runtime-enforcer/templates/agent/daemonset.yaml
@@ -92,8 +92,8 @@ spec:
           httpGet:
             path: /readyz
             port: 8081
-          # we give 45s (5s delay + 8 attempts every 5s) for the startup probe to succeed
-          initialDelaySeconds: 5
+          # we give 50s (10s delay + 8 attempts every 5s) for the startup probe to succeed
+          initialDelaySeconds: 10
           periodSeconds: 5
           failureThreshold: 8
         resources: {{- toYaml .Values.agent.resources | nindent 10 }}
--- a/internal/workloadpolicyhandler/workloadpolicy_handler.go
+++ b/internal/workloadpolicyhandler/workloadpolicy_handler.go
@@ -111,6 +111,10 @@ func (r *WorkloadPolicyHandler) HasSynced(ctx context.Context) error {
 	return nil
 }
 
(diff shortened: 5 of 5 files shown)


Andreagit97:
Instead of rate-limting them, what about turning all the logs into `INFO` (so no more errors) and then when the ctrlmgr stops we simply log a final error

```go
	logger.InfoContext(ctx, "starting manager")
	if err = ctrlMgr.Start(ctx); err != nil {
		if !resolver.IsNRISynchronized() || !wpHandler.IsSynchronized() {
			logger.ErrorContext(ctx, "agent terminated before startup synchronization completed",
				"nriSynchronized", resolver.IsNRISynchronized(),
				"workloadPoliciesSynchronized", wpHandler.IsSynchronized(),
			)
		}
		return fmt.Errorf("failed to start manager: %w", err)
	}
```

In this way we don't log warn or errors because it is an expected behavior but we still return an error in case the startup failed. WDYT?

kyledong-suse:
@Andreagit97, thanks for the suggestion.
That's a good alternative, but my concern is for those warnings like "failed to add pod container from NRI" and "NRI handler has not yet synchronized", do we still want to keep them as WARN or INFO? If we put them into INFO, I'll suggest to rephrase it to make it sounds like an info message instead of a warning.

Andreagit97:
> If we put them into INFO, I'll suggest to rephrase it to make it sounds like an info message instead of a warning.

yep this was the idea. WDYT?

holyspectral:
In addition to what @Andreagit97 said, I want to point out that some of these errors could still be valuable for us to detect performance issues and for users to take actions.

Currently the initial delay of agent's startupProbe is hardcoded as [5 seconds](https://github.com/rancher-sandbox/runtime-enforcer/blob/0b65d03c2bdf4dcccd8a516275361bf83b132ce5/charts/runtime-enforcer/templates/agent/daemonset.yaml#L96).  Perhaps we can increase it to like 10 seconds, so we have less noise in the default scenario while keeping the warning/error log level for later.

Then we can add a case in e2e tests to make sure that no error logs generated by default. 

WDYT? 

On the other hand, it's probably a separate topic, but the `startupProbe` is not configurable right now.  I think maybe it's a good timing to make it configurable, so that users having these error/warning logs, e.g., from a slow cluster or a big cluster, can change it to reduce the noise.

holyspectral:
For the e2e test part, I created https://github.com/rancher-sandbox/runtime-enforcer/issues/606 to follow up.

kyledong-suse:
After another thinking, I think we can combine our ideas.
I propose we make the per-probe logs rate-limited only for WARN(no errors), so we don’t spam logs on startups. Add a single ERROR when `ctrlMgr.Start()` returns and startup sync was not completed.
I don't have strong opinions but we can also additionally bump the `startupProbe.initialDelaySeconds` to 10s or something else.
WDYT? @Andreagit97 @holyspectral

Andreagit97:
Please note that here i'm just referring to these 2 logs

```json
{"time":"2026-04-23T14:26:07.927529398Z","level":"WARN","msg":"NRI handler has not yet synchronized","component":"agent","component":"resolver"}
{"time":"2026-04-23T14:26:07.934065725Z","level":"ERROR","msg":"WorkloadPolicy handler is not synced","component":"agent","error":"failed to list WorkloadPolicies during HasSynced check: Timeout: failed waiting for *v1alpha1.WorkloadPolicy Informer to sync"}
```

IMO they should both become INFO and probably not rate-limited (we can probably raise the delay to 10s). We could add the rate-limiting but it seems probably a little bit too much for now if we turn these logs to info. One log every 10s it means 6 logs in 1 minutes, that seems reasonable, WDYT?

kyledong-suse:
I agree that with rate-limited only for those 2 logs is a bit too much. But I think turning it into INFO makes regressions easier to miss.
What about we change the ERROR to WARN and keeps those logs as WARN, then increase the delay to 10s?

Andreagit97:
> But I think turning it into INFO makes regressions easier to miss.

it's not clear to me what you mean here :thinking:

kyledong-suse:
@Andreagit97 I mean if we modify WARN to INFO, if a startup which takes longer than usual, it might be acceptable for user. However, for our QA or performance point of view, I think WARN is still valuable and we shouldn't make it silent.

Andreagit97:
>I mean if we modify WARN to INFO, if a startup which takes longer than usual, it might be acceptable for user.

yeah the point it that really depends on the number of pods/policies in the cluster so it's hard to say it won't happen by default in some customer setup, if possible i would avoid to alert users where not necessary :thinking:

cmd/agent/main.go
```diff
@@ -95,25 +95,30 @@ func setupWorkloadPolicyHandler(
...
 	}
 	// controller-runtime doesn't support a separate startup probe, so we use the readiness probe instead.
 	// See https://github.com/kubernetes-sigs/controller-runtime/issues/2644 for more details.
 	if err = ctrlMgr.AddReadyzCheck("policy readyz", func(req *http.Request) error {
 		if syncErr := wpHandler.HasSynced(req.Context()); syncErr != nil {
-			logger.ErrorContext(req.Context(), "WorkloadPolicy handler is not synced", "error", syncErr)
-			return fmt.Errorf("WorkloadPolicy handler is not synced: %w", syncErr)
+			logger.InfoContext(
+				req.Context(),
+				"waiting for WorkloadPolicy synchronization to complete",
```
andreaterzolo: i believe this could become a const and be reused also for the error, WDYT?

```go
const msg = "waiting for WorkloadPolicy synchronization to complete"
```

internal/nri/plugin.go
```diff
@@ -163,8 +163,8 @@ func (p *plugin) Synchronize(
 		)
 		if err := p.resolver.AddPodContainerFromNri(podData); err != nil {
 			// This could be recoverable. Returning an error so we can retry.
-			podLogger.ErrorContext(ctx, "failed to add pod container from NRI", "error", err)
-			return nil, fmt.Errorf("failed to add pod container from NRI: %w", err)
+			podLogger.InfoContext(ctx, "NRI pod/container sync not ready yet, will retry", "error", err)
```
andreaterzolo: Same here `"NRI pod/container sync not ready yet, will retry"` could become a const

internal/resolver/nri_api.go
```diff
@@ -113,10 +113,14 @@ func (r *Resolver) NRISynchronized() {
...
 
+func (r *Resolver) IsNRISynchronized() bool {
+	return r.nriSynchronized.Load()
+}
+
 func (r *Resolver) Ping(req *http.Request) error {
 	if !r.nriSynchronized.Load() {
-		r.logger.WarnContext(req.Context(), "NRI handler has not yet synchronized")
-		return errors.New("NRI handler has not yet synchronized")
+		r.logger.InfoContext(req.Context(), "waiting for NRI synchronization to complete")
```
andreaterzolo: same here we can have a const

Andreagit97 (commented):
Thanks!

dottorblaster (approved):
Plus one on @Andreagit97's comments, then we're ready to merge @kyledong-suse, thanks!

Andreagit97 (approved):

PR #611, opened 2026-05-06
test(e2e): switch from Ubuntu to openSUSE BCI images

dottorblaster:

Switching to openSUSE BCIs for the e2e tests. This way we don't have to pin Ubuntu images and rely onto something in which we get to a little bit more of a saying on.

### DIFF
--- a/docs/yaml/opensuse-deployment.yaml
+++ b/docs/yaml/opensuse-deployment.yaml
@@ -1,25 +1,21 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: ubuntu-deployment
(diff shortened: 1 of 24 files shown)


kyledong-suse (commented):
If we want to switch all ubuntu image to bci image, it's better to change those containers name and those testdata files name as well.

holyspectral (approved):
LGTM.  There are some ubuntu references in internal/ folder, but feel free to address them in the next PR.  [ubuntu.txt](https://github.com/user-attachments/files/28030458/ubuntu.txt)

kyledong-suse (approved):

PR #612, opened 2026-05-07
fix: pin ubuntu version to 24.04

kyledong-suse:

Pin the ubuntu version to 24.04

This is because ubuntu 26.04 switched to the rust-version of coreutils and so the binary paths are changed:
* /bin/true => /bin/gnutrue
* /usr/bin/echo => /usr/lib/cargo/bin/coreutils/echo
* /usr/bin/ls => /usr/lib/cargo/bin/coreutils/ls
* /usr/bin/sleep => /usr/lib/cargo/bin/coreutils/sleep
* /usr/bin/cat => /usr/lib/cargo/bin/coreutils/cat

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/troubleshooting.adoc
+++ b/docs/troubleshooting.adoc
@@ -137,7 +137,7 @@ spec:
     spec:
       containers:
       - name: ubuntu
-        image: ubuntu
+        image: ubuntu:24.04
         stdin: true
         tty: true
 ----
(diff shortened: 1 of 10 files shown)


dottorblaster (approved):
Thank you ❤️

PR #620, opened 2026-05-08
docs: add resource consumption benchmark and documentation

holyspectral:

This PR now includes the resource utilization document.  The script itself will be moved in a separate PR.

I will appreciate any feedback!

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/resource-consumption.md
+++ b/docs/resource-consumption.md
@@ -56,17 +56,17 @@ Based on these and the benchmark results, here is our recommendation.  Please co
 
 | #Workload + #WorkloadPolicies | Agent CPU Request/Limit (m) | Agent Memory Request/Limit (Mi) |
 | ----------------------------- | --------------------------- | ------------------------------- |
-| <2000 (small)                 | 500                         | 256                             |
-| 2000~4000 (medium)            | 500                         | 512                             |
-| >4000 (large)                 | 500 or more                 | 768 or more                     |
+| <2000 (small)                 | 250                         | 256                             |
+| 2000~4000 (medium)            | 250                         | 512                             |
+| >4000 (large)                 | 250 or more                 | 768 or more                     |
 
 ### Controller
 
 | #Workload + #WorkloadPolicies | Controller CPU Request/Limit (m) | Controller Memory Request/Limit (Mi) |
 | ----------------------------- | -------------------------------- | ------------------------------------ |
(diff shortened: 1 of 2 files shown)


docs/resource-consumption.md
```diff
@@ -0,0 +1,136 @@
...
+
+In addition to these, Runtime Enforcer agent/controller pods need extra resources to handle security events and behavior learning. They also have to handle bursts during policy and workload reconciliation. 
+
+Based on these and the benchmark results, here is our recommendation.  Please consider this as a starting point for your optimization instead of the final configuration, since every cluster could vary.
+
+### Agent
+
+| #Workload + #WorkloadPolicies | Agent CPU Request/Limit (m) | Agent Memory Request/Limit (Mi) |
+| ----------------------------- | --------------------------- | ------------------------------- |
+| <2000 (small)                 | 500                         | 256                             |
```
andreaterzolo: The memory limit seems inline with what we experienced ourselves running the runtime enforcer and with what you reported in the below benchs. On the other side the CPU usage seems pretty far from what you found, in your benchmarks it is always around 2/3 mCPU and also from what i can see from the demo-dev cluster, do we really expect this big CPU usage?

Andreagit97 (commented):
thank you! Just one comment about the CPU usage

docs/resource-consumption.md
holyspectral: Good question!  I agree that the CPU usage of `500m` specified in our default could be a bit too much.  That's also why I basically keep it the same in all the scenarios in this doc.  The CPU utilization for policy enforcement will go to the cgroup where the user container is running, so that can also explain why we don't need much CPU resource.  

Moreover, I had to increase the node size in my performance test because `500m` is a bit too much for a node with 1 or 2 vCPU, so I believe lower CPU limit would have its benefit as well.  

However, I don't think we have done sufficient tests to justify the lower CPU limit than what we have currently.  Perhaps we can make it like `200m` or `250m` after 0.6.0 and give it some time for us to detect issues.  WDYT?

dottorblaster (approved):
LGTM, just lower the CPU resource limit 👍

Andreagit97 (approved):
thank you! in the future i think we can also lower it to something like 50/100 mcpu, in the end the numbers we are seeing are really low

PR #624, opened 2026-05-12
test: detect excessive error logs

kyledong-suse:

Detect error-level logs in CI by parsing runtime enforcer container logs.

fixes #606 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: "go.mod"
-      - name: Install system dependencies for eBPF build
+      - name: Install system dependencies
         run: |
           sudo apt-get update
           sudo apt-get install -y \
@@ -32,10 +32,15 @@ jobs:
             libelf-dev \
             clang \
             llvm \
-            libbpf-dev
+            libbpf-dev \
+            jq
       - name: Running Test e2e
         run: |
           make test-e2e
+      - name: Check runtime-enforcer logs for unexpected ERROR logs
+        run: |
+          chmod +x hack/ci/check-runtime-enforcer-error-logs.sh
+          hack/ci/check-runtime-enforcer-error-logs.sh test/e2e/logs
       - name: Upload cluster logs
         if: failure()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/hack/ci/check-runtime-enforcer-error-logs.sh
+++ b/hack/ci/check-runtime-enforcer-error-logs.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+LOG_DIR="${1:-test/e2e/logs}"
+NAMESPACE="${RUNTIME_ENFORCER_E2E_NAMESPACE:-run-enf-e2e-runtime-enforcer}"
+RUNTIME_ENFORCER_POD_PREFIX="${RUNTIME_ENFORCER_POD_PREFIX:-runtime-enforcer}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+WHITELIST_JSON="${WHITELIST_JSON:-${SCRIPT_DIR}/whitelist.json}"
+
+if ! command -v jq >/dev/null 2>&1; then
+  echo "jq is required but was not found on PATH" >&2
+  exit 2
+fi
+
+if [[ ! -f "$WHITELIST_JSON" ]]; then
+  echo "whitelist file not found: $WHITELIST_JSON" >&2
+  exit 2
+fi
+
+if [[ ! -d "$LOG_DIR" ]]; then
+  echo "log directory not found: $LOG_DIR" >&2
+  exit 2
+fi
+
+# kind export logs writes per-node folders with a containers/ subtree. Filenames
+# look like "<pod>_<namespace>_<container>-<imageid>.log". Only runtime-enforcer
+# pods (agent, controller) in the e2e namespace.
+mapfile -t log_files < <(
+  find "$LOG_DIR" -type f \
+    \( -name '*.log' -o -name '*.txt' \) \
+    \( \
+      -path "*/containers/${RUNTIME_ENFORCER_POD_PREFIX}-agent-*_${NAMESPACE}_*.log" \
+      -o -path "*/containers/${RUNTIME_ENFORCER_POD_PREFIX}-controller-manager-*_${NAMESPACE}_*.log" \
+    \) \
+  | sort
+)
+
+if (( ${#log_files[@]} == 0 )); then
+  echo "no runtime-enforcer agent or controller log files found under $LOG_DIR for namespace $NAMESPACE" >&2
+  exit 2
+fi
+
+# Full logs remain under LOG_DIR for debugging.
+unexpected="$(
+  jq -Rn --slurpfile whitelist "$WHITELIST_JSON" '
+    ($whitelist[0]) as $entries |
+    [
+      inputs
+      | sub("^[^\\{]*"; "")
+      | fromjson
+      | select((.level? // "" | ascii_upcase) as $l | $l == "ERROR" or $l == "WARN")
+      | select(
+          (.level? // "" | ascii_upcase) as $lvl
+          | (.msg // "") as $m
+          | (.component // "") as $c
+          | ($entries
+            | any(
+                .msg == $m
+                and .component == $c
+                and ((.level // "") | ascii_upcase) == $lvl
+              )
+            | not)
+        )
+    ]
+  ' "${log_files[@]}"
+)"
+
+if jq -e 'length > 0' <<<"$unexpected" >/dev/null; then
+  echo "unexpected ERROR or WARN logs detected in runtime-enforcer logs:" >&2
+  jq . <<<"$unexpected" >&2
+  exit 1
+fi
+
+echo "no unexpected ERROR or WARN logs found"
--- a/hack/ci/whitelist.json
+++ b/hack/ci/whitelist.json
@@ -1,0 +1,27 @@
+[
+  {
+    "level": "ERROR",
+    "msg": "policy deletion denied: policy is still used by pods",
+    "component": "controller"
+  },
+  {
+    "level": "ERROR",
+    "msg": "failed to get policies status",
+    "component": "controller"
+  },
+  {
+    "level": "WARN",
+    "msg": "error during NRI plugin execution, retrying...",
+    "component": "nri-handler"
+  },
+  {
+    "level": "WARN",
+    "msg": "violation buffer full, oldest entry dropped",
+    "component": "agent"
+  },
+  {
+    "level": "WARN",
+    "msg": "logs suppressed by rate limiting",
+    "component": "agent"
+  }
+]


holyspectral (commented):
Thanks!  Just some high-level comments:

1. If we can see the error message along with the error count, that will be great.
2. Not sure if we could use a keyword filter instead of checking the error count.  This way the test will be more reliable if some errors are printed indefinite times. 
3. If we switch to a keyword filter, maybe we could use `-e` in jq to change exit code depending on whether the query finds anything, then we can do (1) at the same time.

.github/workflows/test-e2e.yml
```diff
@@ -33,9 +33,20 @@ jobs:
...
             llvm \
             libbpf-dev
+      - name: Install jq
+        run: |
+          sudo apt-get install -y jq
       - name: Running Test e2e
         run: |
           make test-e2e
+      - name: Fail on unexpected runtime-enforcer error logs
+        if: always()
```
andreaterzolo: i'm also ok with running it only in case of success, in the end i expect errors in case of failure

.github/workflows/test-e2e.yml
```diff
@@ -33,9 +33,20 @@ jobs:
             clang \
             llvm \
             libbpf-dev
+      - name: Install jq
```
andreaterzolo: i would add jq in the list above

hack/ci/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,61 @@
...
+  exit 2
+fi
+
+# kind export logs writes per-node folders with a containers/ subtree. Filenames
+# look like "<pod>_<namespace>_<container>-<imageid>.log". We only match that
+# namespace to avoid false positives from other workloads.
+mapfile -t log_files < <(
+  find "$LOG_DIR" -type f \
+    \( -name '*.log' -o -name '*.txt' \) \
+    -path "*/containers/*_${NAMESPACE}_*.log" \
```
andreaterzolo: great so we are checking both the agent and the controller

Andreagit97 (commented):
thanks!

.github/workflows/test-e2e.yml
kyledong-suse: I was thinking to add it in the above list, but seems the name is for eBPF only.
I'll change the name to be more generic and add install jq there.

kyledong-suse:
Thanks for those good comments @holyspectral and @Andreagit97 
I have made changes to not only fail with error count, but also list those unexpected errors. Currently, we have 3 known error messages
```
{"time":"2026-05-12T23:38:51.268582458Z","level":"ERROR","msg":"policy deletion denied: policy is still used by pods","component":"controller","webhookGroup":"security.rancher.io","webhookKind":"WorkloadPolicy","WorkloadPolicy":{"name":"referenced-wp","namespace":"run-enf-e2e-0ff32245c229ace0698b"},"namespace":"run-enf-e2e-0ff32245c229ace0698b","name":"referenced-wp","resource":{"group":"security.rancher.io","version":"v1alpha1","resource":"workloadpolicies"},"user":"kubernetes-admin","requestID":"b2bd028f-34c4-43e4-9ed9-c391a3e8d5b8","logger":"admission","policyName":"referenced-wp","podCount":1,"podNames":["referenced-wp-pod"]}
{"time":"2026-05-12T23:46:37.396008743Z","level":"ERROR","msg":"failed to get policies status","component":"controller","logger":"WorkloadPolicyStatusSync","err":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.244.0.10:50051: i/o timeout\"","node":"test-controller-e2e-c4569d9f6f82-control-plane"}
{"time":"2026-05-12T23:47:04.676856614Z","level":"ERROR","msg":"Failed to list pod cache","component":"debugger","node":"test-controller-e2e-c4569d9f6f82-control-plane","error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.244.0.10:50051: i/o timeout\""}
```
I created a whitelist for them, and will also print those whitelisted errors even this ci step is passed.

hack/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+
+if (( ${#log_files[@]} == 0 )); then
+  echo "no runtime-enforcer container log files found under $LOG_DIR for namespace $NAMESPACE" >&2
+  exit 2
+fi
+
+tmp_raw="$(mktemp)"
+tmp="$(mktemp)"
+tmp_whitelist="$(mktemp)"
+trap 'rm -f "$tmp_raw" "$tmp" "$tmp_whitelist"' EXIT
```
holyspectral: Actually I think leaving these temp files will be fine.  They will be purged after reboot and they will be helpful for debugging when running in local.

hack/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+)
+
+if (( ${#log_files[@]} == 0 )); then
+  echo "no runtime-enforcer container log files found under $LOG_DIR for namespace $NAMESPACE" >&2
+  exit 2
+fi
+
+tmp_raw="$(mktemp)"
+tmp="$(mktemp)"
+tmp_whitelist="$(mktemp)"
```
holyspectral: nit: maybe add some suffixes, e.g., `$(mktemp --suffix=-whitelist)` so it's easier to identify the temp files?

hack/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+    [
+      (.msg // ""),
+      (.component // ""),
+      (.logger // ""),
+      (.namespace // ""),
+      (.name // ""),
+      (.node // ""),
+      (.err | tostring),
+      (.error | tostring)
+    ]
```
holyspectral: I'm actually fine without the `unique_by` since the number of errors are supposed to be low anyway.  Most importantly it makes the query short and easier to understand.

hack/ci/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+  cat "$tmp_whitelist" >&2
+fi
+
+error_count="$(wc -l <"$tmp" | tr -d "[:space:]")"
+if (( error_count > 0 )); then
+  echo "unexpected ERROR-level logs (gated): ${error_count} (max allowed: ${MAX_ERROR_LOGS})" >&2
+  cat "$tmp" >&2
+fi
+
+if ! jq -en --argjson n "${error_count}" --argjson max "${MAX_ERROR_LOGS}" '$n <= $max' >/dev/null; then
```
holyspectral: Given that we use `unique_by`, the error count here is basically the number of error types, instead of the number of error logs.  I'm fine with both, just want to point out the difference here.

hack/ci/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+($wl_msgs | split(":::") | map(select(length > 0) | split("|||") | select(length == 2) | {msg: .[0], component: .[1]})) as $whitelist |
+($filter) as $fl |
+[inputs]
+| map(select(type == "object"))
+| map(select(
+    (.msg // "") as $m | (.component // "") as $c
+    | if $fl == "exclude" then ($whitelist | any(.msg == $m and .component == $c) | not)
+      elif $fl == "include" then ($whitelist | any(.msg == $m and .component == $c))
+      else false end
+    ))
```
holyspectral: I understand that you want the include/exclude cases to share code, but I'd just define two different queries because it makes the query significantly simpler.  WDYT?

hack/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+  # kind/kubelet logs are CRI-formatted: "<ts> <stream> <F|P> <payload>".
+  # Application output is JSON starting at the first "{".
+  sub("^[^\\{]*"; "") |
+  fromjson? |
+  select((.level? // "" | ascii_upcase) == "ERROR")
+' "${log_files[@]}" >"$tmp_raw"
+
+# $filter: exclude = non-whitelisted (gate); include = whitelisted only (debug).
+_jq_filter_errors_by_whitelist=$'
+($wl_msgs | split(":::") | map(select(length > 0) | split("|||") | select(length == 2) | {msg: .[0], component: .[1]})) as $whitelist |
```
holyspectral: Maybe you could define the whitelist in a separate json file?  So you don't have to convert in bash or jq.  E.g., `jq --rawfile whitelist whitelist.json` or `--slurpfile`

holyspectral (commented):
Generally lgtm.  I also tested in my local and it works great.  Some comments:

1. I think we can simplify the jq query a little bit, so it's easier to understand/maintain.
2. And we can move the whitelist to a separate file.  This way we can get rid of the conversion logic in the script.

hack/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,117 @@
...
+fi
+
+# kind export logs writes per-node folders with a containers/ subtree. Filenames
+# look like "<pod>_<namespace>_<container>-<imageid>.log". We only match that
+# namespace to avoid false positives from other workloads.
+mapfile -t log_files < <(
+  find "$LOG_DIR" -type f \
+    \( -name '*.log' -o -name '*.txt' \) \
+    -path "*/containers/*_${NAMESPACE}_*.log" \
+    2>/dev/null \
```
dottorblaster:
```suggestion
```

I would drop this honestly, with `errexit/pipefail` the pipeline already protects you from a hard failure, suppressing stderr just hides operator-actionable problems on CI

dottorblaster (changes requested):
Thank you so much! One tiny change, then the stuff that @Andreagit97 and @holyspectral commented, then we can merge 👍

hack/check-runtime-enforcer-error-logs.sh
kyledong-suse: I agree, we can remove this dedup and keep jq command simple.

hack/check-runtime-enforcer-error-logs.sh
kyledong-suse: It's a good suggestion. I'll move this script to /hack/ci/ then create a separate JSON file to maintain whitelisted errors.

hack/ci/check-runtime-enforcer-error-logs.sh
kyledong-suse: I agree that 2 separate queries will be easier to read, but those 2 queries share pretty similar logic and code. For now, I still prefer to keep the filter to not maintain 2 nearly identical blocks of code.

.github/workflows/test-e2e.yml
```diff
@@ -24,18 +24,26 @@ jobs:
...
-            libbpf-dev
+            libbpf-dev \
+            jq
       - name: Running Test e2e
         run: |
           make test-e2e
+      - name: Check runtime-enforcer logs for unexpected ERROR logs
+        env:
+          # Keep this configurable so we can allow known errors temporarily.
+          MAX_ERROR_LOGS: "0"
```
andreaterzolo: do we still need this env var?

hack/ci/whitelist.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "msg": "policy deletion denied: policy is still used by pods",
```
andreaterzolo, line 4: on one side i believe this is normal because we are trying to force this error in tests, on the other side i'm not sure the other 2 errors are expected. Is there a transitory situation in which a node is not available?

hack/ci/check-runtime-enforcer-error-logs.sh
```diff
@@ -0,0 +1,99 @@
...
+
+tmp_raw="$(mktemp --suffix=-runtime-enforcer-errors-raw)"
+tmp_unexpected="$(mktemp --suffix=-runtime-enforcer-errors-unexpected)"
+tmp_whitelist="$(mktemp --suffix=-runtime-enforcer-errors-whitelist)"
+
+jq -Rc '
+  # kind/kubelet logs are CRI-formatted: "<ts> <stream> <F|P> <payload>".
+  # Application output is JSON starting at the first "{".
+  sub("^[^\\{]*"; "") |
+  fromjson? |
```
andreaterzolo: why do we need the `?`, do we have non json lines?
```suggestion
  fromjson |
```

Andreagit97 (commented):
I had a look at the script and I think it can be simplified quite a bit while keeping the same goal. Proposal:

* Drop MAX_ERROR_LOGS and the threshold-based gate: any non-whitelisted log ⇒ fail. The threshold risks hiding regressions.
* Drop the "include" branch that prints whitelisted entries: we already have the file with the whitelist
* Collapse the parametric jq filter (exclude/include) and the wrapper function into a single inline jq pipeline, removing the temp files.
* Extend the level match from just ERROR to ERROR + WARN (case-insensitive), so we also catch suspicious warnings.
* Pretty-print output instead of JSON-lines for readability in CI logs.

The script could be something like this
```bash
#!/usr/bin/env bash

set -o errexit
set -o nounset
set -o pipefail

LOG_DIR="${1:-test/e2e/logs}"
NAMESPACE="${RUNTIME_ENFORCER_E2E_NAMESPACE:-run-enf-e2e-runtime-enforcer}"

SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
WHITELIST_JSON="${WHITELIST_JSON:-${SCRIPT_DIR}/whitelist.json}"

if ! command -v jq >/dev/null 2>&1; then
  echo "jq is required but was not found on PATH" >&2
  exit 2
fi

if [[ ! -f "$WHITELIST_JSON" ]]; then
  echo "whitelist file not found: $WHITELIST_JSON" >&2
  exit 2
fi

if [[ ! -d "$LOG_DIR" ]]; then
  echo "log directory not found: $LOG_DIR" >&2
  exit 2
fi

# kind export logs writes per-node folders with a containers/ subtree. Filenames
# look like "<pod>_<namespace>_<container>-<imageid>.log". We only match that
# namespace to avoid false positives from other workloads.
mapfile -t log_files < <(
  find "$LOG_DIR" -type f -name '*.log' \
    -path "*/containers/*_${NAMESPACE}_*.log" | sort
)

if (( ${#log_files[@]} == 0 )); then
  echo "no runtime-enforcer container log files found under $LOG_DIR for namespace $NAMESPACE" >&2
  exit 2
fi

unexpected="$(
  jq -R --slurpfile wl "$WHITELIST_JSON" '
    sub("^[^\\{]*"; "") | fromjson?
    | select((.level? // "" | ascii_upcase) as $l
             | $l == "ERROR" or $l == "WARN")
    | . as $e
    | select(($wl[0] | any(.msg == ($e.msg // "")
                           and .component == ($e.component // ""))) | not)
  ' "${log_files[@]}"
)"

if [[ -n "$unexpected" ]]; then
  echo "unexpected ERROR/WARN logs from runtime-enforcer:" >&2
  printf '%s\n' "$unexpected" >&2
  exit 1
fi
```

It is AI generated so we need to double check it really works :/ WDYT?

hack/ci/check-runtime-enforcer-error-logs.sh
kyledong-suse: It's a defensive coding, just in case the log file contains non-json lines.

hack/ci/whitelist.json
kyledong-suse: I think the other 2 are either timing or connection issues we should investigate later.

.github/workflows/test-e2e.yml
kyledong-suse: removed

hack/ci/check-runtime-enforcer-error-logs.sh
andreaterzolo: uhm i see but here we are in our e2e tests so probably it's better to fail and let us know that we have some lines that are not json formatted. If we have some non-json lines we should fix them IMO

kyledong-suse: Actually another concern for this is we're scanning runtime-enforcer agent, controller, debugger and otel container logs in the test/e2e/logs. if we only scan agent and controller logs, we can focus on json format only. WDYT?

andreaterzolo: oh i see, great point. yep we should probably just focus on the agent and the controller for now and remove the `?`, in the end this is what we are already doing ignoring not json logs, i assume the debugger and the otel collector don't have json logs so we are just implicitly ignoring them

Andreagit97 (approved):

dottorblaster (approved):

PR #630, opened 2026-05-18
fix: add -r flag to jq in SBOM digest extraction

dottorblaster:

Without -r, jq emits a JSON-encoded string with literal quotes around the digest, which get propagated through `$GITHUB_OUTPUT` into the downstream `crane blob` call. This PR fixes it.

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Andreagit97 (approved):
thanks

PR #665, opened 2026-06-03
chore: rename ubuntu identifiers to opensuse across tests and docs

dottorblaster:

Just renaming the leftovers to openSUSE for consistency :-)

fixes #663 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/podworkload/workload.go
+++ b/internal/podworkload/workload.go
@@ -36,23 +36,23 @@ func parseDeployment(podName, templateHash string) (string, workloadkind.Kind) {
 	//
 	// Concrete examples:
 	// 1. Regular case
-	//    deployment-name: ubuntu-deployment-674bcc58f4-pwvps
-	//    pod-name: ubuntu-deployment
+	//    deployment-name: opensuse-deployment-674bcc58f4-pwvps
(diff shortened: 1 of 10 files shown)


internal/podworkload/workload.go
andreaterzolo, line 1: This is the unique file that shouldn't receive any rename. Please note there are some examples based on the len 

```
// 2. 63 characters
	//    deployment-name: opensuse-deploymentttttttttttttttttttttttttttttttttttttttttttt
	//    pod-name: opensuse-deploymenttttttttttttttttttttttttttttttttttttttttq8fcg
	//    In this case we have just the final [random(q8fcg)] without the `-`. The name of the deployment is truncated.
```

since `opensuse` != `ubuntu` we cannot just rename since it will cause a wrong example. I would suggest to keep ubuntu here

internal/podworkload/workload_test.go
andreaterzolo, line 1: also in this file i would probable leave `ubuntu` we are not deploying anything just using `ubuntu` as a simple name

Andreagit97 (approved):

PR #678, opened 2026-06-10
docs(rfc): violation liveness and resolution tracking

dottorblaster:

Just wrote the RFC to implement the concept of "active" violations.

fixes #672

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/rfc/0008-violation-liveness.md
+++ b/docs/rfc/0008-violation-liveness.md
@@ -1,0 +1,487 @@
+|              |                                                              |
+| :----------- | :----------------------------------------------------------- |
+| Feature Name | Violation Liveness and Resolution Tracking                   |
+| Start Date   | 2026-06-10                                                   |
+| Category     | Security / Observability                                     |
+| RFC PR       | https://github.com/rancher-sandbox/runtime-enforcer/pull/678 |
+| State        | **ACCEPTED**                                                 |
+
+# Summary
+
+[summary]: #summary
+
+Right now violations are a flat historical log: once recorded they stay in
+`status.violations` until the 100-record cap pushes them out. There is no way
+to tell which ones still represent an active risk and which have already been
+resolved by an allowlist update or a workload redeployment, and operators lack
+context about which workload produced each violation. We enrich violation
+records with workload metadata, support punctual violation acknowledgements
+via a controller-consumed policy annotation, and delete expired entries when
+building the violations list. This gives operators meaningful context and a
+clean way to acknowledge known issues while automatically removing resolved
+ones.
+
+# Motivation
+
+[motivation]: #motivation
+
+Without liveness, every violation looks equally urgent. A security engineer
+triaging a policy with 40 violations has no way to know that 35 of them came
+from pods that were redeployed last week, or that 3 more refer to binaries
+that have since been allowlisted. Dashboards, alerts, and compliance reports
+all end up noisier than they need to be, and there is no clean signal for
+"is this workload currently clean?"
+
+## Examples
+
+[examples]: #examples
+
+**Executable added to the allowlist.** A violation for `/usr/local/bin/debug-tool`
+sits in the violations list. The operator edits the policy and adds the
+path under `spec.rulesByContainer["app"].executables.allowed`. On the next
+reconcile the controller sees the match and deletes the record from the list.
+
+**Violation acknowledged.** An operator notices a violation for
+`/opt/scripts/cleanup.sh` and knows it is part of an ongoing incident. They
+read the violation's `id` from `status.violations` (e.g. via
+`kubectl describe workloadpolicy`) and patch the policy with an annotation
+referencing it:
+
+```yaml
+metadata:
+  annotations:
+    runtime-enforcer.security.rancher.io/acknowledge/42: "ongoing incident, see JIRA-1234"
+```
+
+On the next reconcile the controller:
+
+1. Finds the annotation, extracts the violation `id` and `reason`.
+2. Looks up the matching record in `status.violations` and removes it.
+3. Appends it to `status.acknowledgedViolations` (carrying the original
+   record, the reason, and the acknowledgement timestamp).
+4. Emits an OTEL log with the violation data and the reason.
+5. Removes the annotation from the policy.
+
+After this, `status.violations` only contains items that actually need
+attention, and `status.acknowledgedViolations` is the audit trail of
+explicitly triaged ones.
+
+**Mixed state.** After a few days, only unresolved violations remain in the
+list:
+
+```
+status.violations:
+  /usr/bin/nc                 (still running, not allowed, still in list)
+
+status.acknowledgedViolations:
+  /opt/scripts/cleanup.sh     (acknowledged via annotation, moved here)
+```
+
+The violations list now contains only items that actually need attention.
+
+# Detailed design
+
+[design]: #detailed-design
+
+## Violation record enrichment
+
+Each `ViolationRecord` gains a stable `id` plus workload fields
+(`workloadName`, `workloadKind`) alongside the existing `podName`. The
+`id` makes a single violation addressable from outside the controller; the
+workload fields give operators immediate context about which workload
+produced the violation. Workload metadata is populated from the pod's owner
+references when the violation is recorded.
+
+The agent's in-memory dedup key (policy, pod, container, executable, action)
+is reused by the controller to recognise that a violation scraped from the
+agent on a later tick is the same one it already knows about, so the `id`
+(and workload fields) are preserved across scrapes.
+
+## Violation `id` as an evergrowing integer
+
+The `id` is a per-policy, monotonically increasing `uint64` assigned by the
+controller when the record is first added to the status. It is not a UUID
+and not globally unique: it is unique within the policy's violation stream
+and is meant to be a short, human-readable reference ("see violation 42 of
+policy foo/bar"). Ids are never reused, even after a record is
+acknowledged or removed by the allowlist-expiration pass, so they remain
+stable references for OTEL logs, audit trails, and operator
+back-and-forth.
+
+The counter lives on the policy itself as `status.nextViolationID`. The
+controller allocates the next id by reading the current value, writing it
+into the new `ViolationRecord.id`, and bumping `status.nextViolationID` by
+one. On the first reconcile of a fresh policy the field defaults to `1`.
+On upgrade, existing policies with no `nextViolationID` are initialized
+to `max(existingIds) + 1` (or `1` if the list is empty) on the next
+reconcile that adds a new record. We accept that pre-upgrade records
+without an `id` are not addressable, mirroring the existing decision not
+to backfill workload fields.
+
+Using an integer keeps annotation keys short and the wire format
+opaque-free ("`/acknowledge/42`" is much friendlier than a UUID) at the
+cost of the controller having to coordinate the counter through the
+status field. Because status writes go through the API server's
+optimistic-concurrency loop, a double-allocation cannot happen: the
+controller reads the current counter, allocates an id, and writes the
+status with the bumped counter in a single update, retrying on
+conflict.
+
+## Acknowledgement via annotation
+
+Operators acknowledge a specific violation by setting an annotation on the
+`WorkloadPolicy` with the form:
+
+```
+<annotation-prefix>/<violation-id>: "<reason>"
+```
+
+The default annotation prefix is
+`runtime-enforcer.security.rancher.io/acknowledge`. The key's suffix is
+the `id` of the violation being acknowledged, taken from
+`status.violations[].id`. The value is a free-form human-readable reason
+recorded into `status.acknowledgedViolations`.
+
+Example:
+
+```yaml
+metadata:
+  annotations:
+    runtime-enforcer.security.rancher.io/acknowledge/42: "ongoing incident, see JIRA-1234"
+```
+
+On every reconcile the controller:
+
+1. Collects all annotations on the policy whose key has the acknowledgement
+   prefix.
+2. For each such annotation, parses the violation `id` from the key and the
+   `reason` from the value.
+3. Looks up the record in `status.violations` by `id`. If found:
+   - Removes it from `status.violations`.
+   - Appends a new entry to `status.acknowledgedViolations` carrying the
+     full original `ViolationRecord`, the `reason`, and the
+     acknowledgement timestamp.
+   - Emits an OTEL log record with `event_name=policy_violation_acknowledged`
+     containing the violation data and the reason (see OTEL logging below).
+   - Records the annotation key in a "to-be-removed" set.
+4. If the `id` does not match any current violation, the annotation is
+   treated as a no-op (still removed), and a warning OTEL log is emitted
+   so the operator can tell their annotation was stale.
+5. After processing all acknowledgement annotations, the controller removes
+   them from the policy object via a patch. Acknowledgements are
+   one-shot: once consumed, the annotation is gone and the acknowledgement
+   lives on in `status.acknowledgedViolations`.
+
+The acknowledgement is "punctual" in the sense that it targets one specific
+violation by `id` and disappears after being processed. It does not affect
+future violations of the same executable, so a reappearing violation will
+have to be acknowledged again.
+
+## Resolved violations via allowlist update
+
+When the operator edits the policy and adds an executable to
+`spec.rulesByContainer[<container>].executables.allowed`, the controller
+walks `status.violations` and deletes every record whose `executablePath`
+is now allowed for its `containerName`. This is independent of the
+acknowledgement flow above and does not produce an
+`acknowledgedViolations` entry: resolved-by-allowlist is not a triage
+decision, it is an actual remediation, and the audit trail for that is the
+allowlist itself.
+
+## Status layout
+
+```go
+type ViolationRecord struct {
+    ID             uint64       `json:"id"`
+    Timestamp      metav1.Time `json:"timestamp"`
+    PodName        string      `json:"podName"`
+    ContainerName  string      `json:"containerName"`
+    ExecutablePath string      `json:"executablePath"`
+    NodeName       string      `json:"nodeName"`
+    Action         string      `json:"action"`
+    WorkloadName   string      `json:"workloadName,omitempty"`
+    WorkloadKind   string      `json:"workloadKind,omitempty"`
+}
+
+type AcknowledgedViolation struct {
+    Violation      ViolationRecord `json:"violation"`
+    Reason         string          `json:"reason"`
+    AcknowledgedAt metav1.Time     `json:"acknowledgedAt"`
+}
+
+type WorkloadPolicyStatus struct {
+    // ... existing fields ...
+    Violations              []ViolationRecord         `json:"violations,omitempty"`
+    AcknowledgedViolations  []AcknowledgedViolation   `json:"acknowledgedViolations,omitempty"`
+    ActiveViolationCount    int                       `json:"activeViolationCount,omitempty"`
+    NextViolationID         uint64                      `json:"nextViolationID,omitempty"`
+}
+```
+
+`activeViolationCount` is derived from `len(status.violations)` and is the
+field dashboards and alerts should consume. `acknowledgedViolations` is the
+audit trail of explicit triage decisions; it is capped at
+`MaxViolationRecords` (100) entries, the same cap that applies to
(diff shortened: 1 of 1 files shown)


docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,181 @@
...
+The proto, the agent's `ViolationBuffer`, and the OTEL event pipeline are
+untouched. RBAC is already sufficient: the controller has `get/list/watch`
+on pods. On upgrade, existing records without the field default to active
+and get re-evaluated on the first reconcile.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+- Pod name collisions on StatefulSets can briefly mis-report liveness: a new
```
andreaterzolo: uhm this is definitely not ideal, maybe we should add the podUID in the violation record, so that we are sure when a pod is restarted

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,181 @@
...
+  pod with the same name makes an old record look active again. The
+  100-record trim and the fact that a genuinely recurring violation will
+  produce a fresh record with a new timestamp make this tolerable for now.
+- Runtime-dropped binaries are covered eventually (pod is gone, violation expires).
```
andreaterzolo: to be honest this is not super clear to me. If the image is changed the violations should expire since the pod will be redeployed, am i missing something?

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,181 @@
...
+# Detailed design
+
+[design]: #detailed-design
+
+The `ViolationRecord` struct gains one field:
+
+```go
+// active indicates whether the violation is currently live.
+// +optional
+Active bool `json:"active,omitempty"`
```
andreaterzolo: instead of `Active` why don't we change the name of the violation array to `ActiveViolations` and when a violation is no more active we drop it or we move it to another array `ResolvedViolation`? What i don't love is the fact that the violation could change state multiple times while once resolved the violation should stay resolved

andreaterzolo: * let's say we have a violation on  `/usr/bin/curl` 
* the user will add curl because he forgot about it. 

IMO the violation should be resolved and it shouldn't have the possibility to become active again by design. Keeping an active field gives the wrong idea that the violation could become active again, but i don't think this should be the case

docs/rfc/0008-violation-liveness.md
dottorblaster: Yes but for example if you change the image and you keep having an entrypoint that does `wget` of something and that binary is nasty you can incur in new violations. I can strip this off, I just included it for the sake of completeness :smile:

docs/rfc/0008-violation-liveness.md
dottorblaster: What I thought initially was to have a list of `violationExceptions` or something like that, but the concept is very much the same. I agree that an expired violation should stay expired, I'm just trying to wrap my head around this without second guessing myself

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,181 @@
...
+
+```
+/usr/local/bin/debug-tool   active: false   (allowlisted)
+/tmp/malware.sh             active: false   (pod gone)
+/opt/scripts/cleanup.sh     active: true    (still running, not allowed)
+/usr/bin/nc                 active: true    (still running, not allowed)
+```
+
+Filtering on `active == true` gives you the two items that actually need
+attention.
```
flavio: I don't think that marking a resolved the issues that came from a pod that is no longer running is a good idea.

Take the example you made, the `malware.sh` that was blocked now is no longer standing out. IMHO it should be visible for as long as a user decides to silence it.

docs/rfc/0008-violation-liveness.md
dottorblaster: I agree :+1:

docs/rfc/0008-violation-liveness.md
flavio: I like this idea, let's keep that in mind

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,181 @@
...
+
+# Alternatives considered
+
+[alternatives]: #alternatives
+
+## Rejected: violation liveness via image digest
+
+We also considered a design where each `ViolationRecord` would carry the
+container's resolved image **digest** (from
+`.status.containerStatuses[].imageID`, never the mutable tag), captured at
```
flavio, line 431: While having the image digest is better, this makes things more complicated (as pointed out next).

What if we stick to the image tag instead? We can enforce best practices (using shasums) with documentation and with a mutating admission controller (KW adm controller has already a policy for that).

flavio (commented):
I'm not 100% sure about this design, but it moves in the right direction. I think it also handles scenarios like the removal of a Workload or the de-assignment of a policy. In both cases the Pods with violations would be removed by Kubernetes, and their violations would be marked as no longer active.

I wonder if what we currently have isn't too "pod-centric". As a user, I see there's a violation on a Pod, but I will need to know which workload owns it (to further study how to address that). Moreover, I think users tend to look at high order resources (Deployments, StatefulSet,...) and then dig deeper when something is misbehaving.

Should we add a owner reference on the violation, pointing to the workload that owns that?
Should we introduce a new CRD to track violations? I know we want to keep thing simple, but we're soon starting to have a UI developed and they will definitely build on top of our CRDs; hence we should come up with something that can last for some time...

docs/rfc/0008-violation-liveness.md
andreaterzolo: the pod UID should remain stable unless there is a rollout so we won't silence a warning unless the parent workload (deployment, daemonset) is deleted or there is a rollout. If after malware.sh the container restarts the pod UID doesn't change so we shouldn't resolve the violation.
Unfortunately Job/Cronjobs are kind of a corner cases because if there is a violation inside them the and they complete their task, the violation will be automatically solved

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+## Examples
+
+[examples]: #examples
+
+**Executable added to the allowlist.** A violation for `/usr/local/bin/debug-tool`
+sits in the violations list. The operator edits the policy and adds the
+path under `spec.rulesByContainer["app"].executables.allowed`. On the next
+reconcile the controller sees the match and deletes the record from the list.
+
+**Workload redeployed.** A violation references pod `app-7d8f9c-x2k4p` which
```
andreaterzolo: if i recall correctly we don't need anymore this case, there should be only 2 cases:
* the user forgot a binary in the policy -> the user adds it manually and the controller will clean the violations
* there is a real violation -> the user needs to `acknowledged` it and the controller will clean the acknowledged violation.

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+  - id: incident-1--2026-06-12
+    reason: "incident in triage"
+    match:
+      podUID: "8f3c0e3d-..."
+      containerName: "app"
+      executablePath: "/tmp/malware.sh"
+```
+
+When the controller reconciles and builds the violations list, it deletes
+entries that are expired (pod no longer exists, or executable is now
```
andreaterzolo:
```suggestion
entries that are expired (executable is now
```

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+allowlisted) or acknowledged (matches an entry in `spec.acknowledgedViolations`).
+This keeps `status.violations` clean and actionable without requiring operators
+to filter on status fields.
+
+The agent and gRPC proto gain workload fields to pass through to the
+controller. RBAC is already sufficient: the controller has `get/list/watch`
```
andreaterzolo: the agent should already provide the information on the workload through gRPC so the controller shouldn't do anything.

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+container's resolved image **digest** (from
+`.status.containerStatuses[].imageID`, never the mutable tag), captured at
+the time the violation was emitted by the agent. A record would be active
+if and only if both its digest matched what the container is currently
+running and its `executablePath` was not in the container's allowlist. The
+controller would expose a derived `activeViolationCount` and a `Compliant`
+condition (`True` when the count is zero), giving monitors a single field
+to read.
+
+Clearance would be tied to the actualremediation action: rebuilding the image
```
andreaterzolo:
```suggestion
Clearance would be tied to the actual remediation action: rebuilding the image
```

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+
+# Unresolved questions
+
+[unresolved]: #unresolved-questions
+
+- Should we add deduplication of violation records by (workload, container,
+  executable) with a count field? Repeated attacks can produce many similar
+  records, potentially causing unwanted data loss when the 100-record cap is
+  hit. This is orthogonal to liveness and can be added separately.
+- Should the controller expose a derived `activeViolationCount` in status,
```
andreaterzolo: yep this seems a good thing to have

docs/rfc/0008-violation-liveness.md
flavio: yes, I agree with @Andreagit97

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+**Workload redeployed.** A violation references pod `app-7d8f9c-x2k4p` which
+ran a binary dropped at runtime into `/tmp`. The operator does
+`kubectl rollout restart deployment/app`. The old pod is deleted, a new one
+comes up with a different name. The controller's pod informer no longer finds
+`app-7d8f9c-x2k4p` in the cluster, so the violation is deleted from the list.
+The new pod starts with no violations of its own.
+
+**Violation acknowledged.** An operator sees a violation for `/opt/scripts/cleanup.sh`
+and knows it's part of an ongoing incident. They add it to
+`spec.acknowledgedViolations` with a reason and match criteria. On the next
```
flavio: I'm not convinced we should add this information to the `spec` field, since it not about declaring a desired state.

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+The violations list now contains only items that actually need attention.
+
+# Detailed design
+
+[design]: #detailed-design
+
+Each `ViolationRecord` gains workload fields (`workloadName`, `workloadKind`,
+`workloadNamespace`) alongside the existing `podName`, giving operators
```
flavio: I think we can avoid the workloadNamespace, since the policy sits exactly in the same namespace as the workload.

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+immediate context about which workload produced the violation. This is
+populated from the pod's owner reference when the violation is recorded.
+
+The policy spec gains an `acknowledgedViolations` list where operators can
```
flavio: Maybe make that a top level key of the object? I would do a quick search to see if there are some best practices for that.

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+explicitly mark specific violations as known and accepted. Each entry includes
+an identifier, a reason string, and match criteria (pod UID, container name,
+executable path). For example:
+
+```yaml
+spec:
+  acknowledgedViolations:
+  - id: incident-1--2026-06-12
+    reason: "incident in triage"
```
flavio: The user has to come up with `id` and `reason`. I'm fine with reason, I wonder about id, we might want to help the user come up with an automatic value for that field

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+This keeps `status.violations` clean and actionable without requiring operators
+to filter on status fields.
+
+The agent and gRPC proto gain workload fields to pass through to the
+controller. RBAC is already sufficient: the controller has `get/list/watch`
+on pods and can read owner references to populate workload metadata. On
+upgrade, existing records without workload fields are left as-is and still
+subject to expiration checks.
+
+This design prioritizes clarity over history preservation: resolved violations
```
flavio: please extend the RFC by pointing out that for history purpose the user must configure an endpoint to receive all the otel logs we generate

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+were cleared.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+- Deleting expired violations means we lose historical data about past
+  incidents. Operators who want to audit what happened last month will need
+  to rely on external logging or the OTEL event pipeline rather than the
+  policy's status field.
```
flavio: I would move that above

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,202 @@
...
+- Pod name collisions on StatefulSets can cause premature deletion: if a new
+  pod with the same name comes up before the violation is reconciled, the
+  old violation might persist longer than intended. Using pod UIDs in the
+  match criteria mitigates this for acknowledged violations.
```
flavio: good idea, can we make pod UID optional? we could have our kubectl plugin deal with that automatically

docs/rfc/0008-violation-liveness.md
dottorblaster: Sorry, leftover lol

docs/rfc/0008-violation-liveness.md
dottorblaster: Absolutely!

docs/rfc/0008-violation-liveness.md
dottorblaster: Sounds reasonable. Maybe in the `status`?

docs/rfc/0008-violation-liveness.md
dottorblaster: Like the id field but generated? I like it

docs/rfc/0008-tickets.md
andreaterzolo, line 1: uhm this seems a claude leftover 😂

docs/rfc/0008-violation-liveness.md
andreaterzolo: uhm the status doesn't seem the correct place, since it is something controller-owned and shouldn't be touched by the user. The user here is asking the controller to acknowledge the violations so it doesn't seem wrong to put it into the spec, we can see it as a desired state "we want these violations to be acknowledged", and the controller will eventually acknowledge them

docs/rfc/0008-tickets.md
dottorblaster: :sweat_smile: I wanted a skeleton for the implementation phase, also very outdated right now :+1:

docs/rfc/0008-violation-liveness.md
dottorblaster: My two cents: status is out, spec works, the correct answer to this would be a separate resource but I also want to prevent premature CRD proliferation and keep a KISS approach. Folding it into the policy spec means one RBAC grant covers both triage and enforcement, which we'll regret in a security product. I'm down to this but let's keep it self-contained and document the RBAC coupling as known debt.

Andreagit97 (approved):

docs/rfc/0008-violation-liveness.md
flavio: Sorry, I really don't like the usage of `status.something` for that.

What about having an annotation per acknowledged incident?

```yaml
apiVersion: runtime-enforcer.io/v1alpha1
kind: Policy
metadata:
  name: restrict-app-exec
  annotations:
    # Pattern: [prefix]/[violation-id] : [reason]
    ack.runtime-enforcer.io/b2e45a90: "Dev team investigating memory leak, allowed temporary exec"
    ack.runtime-enforcer.io/c8f19b22: "Ops ticket INC-9942 - manual DB migration script"
spec:
```

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+
+The violations list now contains only items that actually need attention.
+
+# Detailed design
+
+[design]: #detailed-design
+
+## Violation record enrichment
+
+Each `ViolationRecord` gains a stable `id` (a UUID generated by the controller
```
andreaterzolo: do you see any advantage in using a specific generated UUID over the incremental counter that we already have in the policy?

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+1. Collects all annotations on the policy whose key has the acknowledgement
+   prefix.
+2. For each such annotation, parses the violation `id` from the key and the
+   `reason` from the value.
+3. Looks up the record in `status.violations` by `id`. If found:
+   - Removes it from `status.violations`.
+   - Appends a new entry to `status.acknowledgedViolations` carrying the
+     full original `ViolationRecord`, the `reason`, and the
+     acknowledgement timestamp.
+   - Emits an OTEL log record with `event_name=policy_violation_acknowledged`
```
andreaterzolo, line 164: not sure if we want the otel log on the acknowledgement in the first iteration

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+     containing the violation data and the reason (see OTEL logging below).
+   - Records the annotation key in a "to-be-removed" set.
+4. If the `id` does not match any current violation, the annotation is
+   treated as a no-op (still removed), and a warning OTEL log is emitted
```
andreaterzolo, line 168: maybe we can start with a warning in the controller

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+5. After processing all acknowledgement annotations, the controller removes
+   them from the policy object via a patch. Acknowledgements are
+   one-shot: once consumed, the annotation is gone and the acknowledgement
+   lives on in `status.acknowledgedViolations`.
+
+The acknowledgement is "punctual" in the sense that it targets one specific
+violation by `id` and disappears after being processed. It does not affect
+future violations of the same executable, so a reappearing violation will
+have to be acknowledged again.
+
```
flavio, line 179: We also have to mention the `status.acknowledgedViolations` is capped to 100 items

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+`acknowledgedViolations` list and the OTEL log. This keeps the
+annotation-based design layered cleanly: the controller speaks
+annotations, the plugin speaks operator UX, the two are kept in sync by
+the annotation contract.
+
+## RBAC and upgrade
+
+The agent and gRPC proto will pass the workload provenance data to the
+controller. RBAC is already sufficient: the controller has `get/list/watch`
+on pods and can read owner references to populate workload metadata. The
```
andreaterzolo: This is not true
```suggestion
controller. The
```

docs/rfc/0008-violation-liveness.md
```diff
@@ -0,0 +1,439 @@
...
+controller needs `update`/`patch` on `workloadpolicies` to remove the
+acknowledgement annotations, which it already has (it writes status today).
+
+On upgrade, existing records without `id` and workload fields are
```
andreaterzolo: i don't think this section is relevant since we are not yet running this software in production

flavio (approved):
LGTM

docs/rfc/0008-violation-liveness.md
```diff
@@ -98,6 +97,36 @@ is reused by the controller to recognise that a violation scraped from the
 agent on a later tick is the same one it already knows about, so the `id`
 (and workload fields) are preserved across scrapes.
 
+## Violation `id` as an evergrowing integer
+
+The `id` is a per-policy, monotonically increasing `int64` assigned by the
```
flavio: we should make that a uint64 to have even more room for growth

docs/rfc/0008-violation-liveness.md
dottorblaster: Just going with a uint64 is ok for now, thanks

flavio (approved):
LGTM

PR #692, opened 2026-06-23
chore(build): set helm.sh/resource-policy via kubebuilder marker

dottorblaster:

Polishing the Makefile and replacing sed with an RBAC annotation :+1: 

fixes #682


Andreagit97 (approved):
cool! thanks

kyledong-suse (approved):
LGTM, thanks!

PR #714, opened 2026-06-30
test(e2e): replace E2E_SKIP_DEPENDENCIES with declarative E2E_DEPENDENCIES

dottorblaster:

Replace the boolean `E2E_SKIP_DEPENDENCIES` toggle with a single allowlist env var `E2E_DEPENDENCIES`.

Semantics:

- unset/empty: install all optional deps (keeps compatibility with the previous default)
- `none`: installs only the runtime-enforcer chart
- `name[,name,...]`: installs only the named deps, case-insensitive

fixes #661

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,94 @@ type helmChart struct {
...
+	if raw == "" {
+		return optionalCharts(), nil
+	}
+	if strings.EqualFold(raw, "none") {
+		return nil, nil
+	}
+
+	all := optionalCharts()
+	known := make(map[string]struct{}, len(all))
+	knownOrder := make([]string, 0, len(all))
```
holyspectral: It feels like `knownOrder` is a little overkill just to format the error.  lol

test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,94 @@ type helmChart struct {
...
+	for _, c := range all {
+		known[strings.ToLower(c.name)] = struct{}{}
+		knownOrder = append(knownOrder, c.name)
+	}
```
holyspectral: Perhaps we don't have to declare `all` since `optionalCharts()` always returning a new slice?
```suggestion
	known := make(map[string]struct{}, len(all))
	knownOrder := make([]string, 0, len(all))
	for _, c := range optionalCharts() {
		known[strings.ToLower(c.name)] = struct{}{}
		knownOrder = append(knownOrder, c.name)
	}
```

test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,94 @@ type helmChart struct {
...
+			return nil, fmt.Errorf(
+				"unknown E2E_DEPENDENCIES name %q; valid names: %s",
+				tok,
+				strings.Join(knownOrder, ", "),
+			)
+		}
+		requested[key] = struct{}{}
+	}
+
+	selected := slices.DeleteFunc(slices.Clone(all), func(c helmChart) bool {
```
holyspectral: Similarly we probably don't need to clone it.  
```suggestion
	selected := slices.DeleteFunc(optionalCharts(), func(c helmChart) bool {
```

holyspectral (commented):
LGTM.  Just some minor comments.

test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,94 @@ type helmChart struct {
 	helmOptions   []helm.Option
 }
 
-func getCharts() []helmChart {
+func optionalCharts() []helmChart {
```
andreaterzolo: probably here i would just return a map[string]helmChart

test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,94 @@ type helmChart struct {
...
+//   - comma-separated names  -> only the named deps, in canonical order
+//
+// Names are matched case-insensitively. An unknown name fails fast with
+// an error listing the valid dependency names.
+func selectOptionalCharts() ([]helmChart, error) {
+	raw := strings.TrimSpace(os.Getenv("E2E_DEPENDENCIES"))
+	if raw == "" {
+		return optionalCharts(), nil
+	}
+	if strings.EqualFold(raw, "none") {
```
andreaterzolo: IMO the logic here should become
```go

	switch raw {
	case "none":
		return charts, nil
	case "":
		for _, chart := range chartsMap {
			charts = append(charts, chart)
		}
	default:
		// get from the map and append the chart if exists, otherwise return an error
	}
```

test/e2e/e2e_suite_test.go
```diff
@@ -58,11 +54,68 @@ type helmChart struct {
...
 	// The order of the charts is relevant because the installation
 	// of certain charts may depend on others being present.
 	//
-	// There are the charts that are always installed by tests.
+	// The runtime-enforcer chart is always installed. Optional
+	// dependencies are selected declaratively via selectOptionalCharts
+	// and prepended so they are installed first.
+	optional := selectOptionalCharts()
+
 	charts := []helmChart{
```
andreaterzolo: nit
```suggestion
	charts := selectOptionalCharts()
	return append(charts, []helmChart{ ...
```

Andreagit97 (approved):

Andreagit97 (approved):

PR #716, opened 2026-07-02
feat: add violation IDs and identify violations

### LINKED ISSUE #694: feat(violations): add stable ids and workload provenance to `ViolationRecord`
`ViolationRecord` today is just a flat historical entry: `podName`, `containerName`, `executablePath`, `nodeName`, `action`, `timestamp`. You can't point at a single violation from outside the  controller, and there's no hint of which workload produced it.          

Three pieces. First, extend ViolationRecord with ID `uint64`, `WorkloadName`, `WorkloadKind`, and add `NextViolationID uint64` to `WorkloadPolicyStatus`. The id is a per-policy monotonic counter allocated by the controller when a record is first added: read `status.nextViolationID`, write it into the record, bump the counter, all in one status update so optimistic concurrency rules out double-allocation. Fresh policy: counter starts at 1.

Second, `WorkloadName`/`WorkloadKind` come from the pod's owner references at recording time.

Third, reuse the agent's in-memory dedup key (policy, pod, container, executable, action) in the controller, so a violation re-scraped on a later tick is recognised as the same record and keeps its id and workload fields instead of being re-allocated.

### Acceptance criteria                                                                                                                                                                 
                                                                                                                                                                                     
 - [ ] `ViolationRecord` carries `id`, `workloadName`, `workloadKind`; `WorkloadPolicyStatus` carries `nextViolationID`.
 - [ ] New violations get a unique per-policy id with the counter bumped atomically; concurrent reconciles of the same policy can't produce duplicate ids (test that fakes a status-update conflict).
 - [ ] Workload name/kind are populated from owner references for new violations, and a re-scraped violation keeps the same id and workload fields (dedup test).

dottorblaster:

Adding stable violation IDs to the `ViolationRecord` as per our last RFC :-)

fixes #694

**Checklist**:

- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/api/v1alpha1/workloadpolicy_types.go
+++ b/api/v1alpha1/workloadpolicy_types.go
@@ -72,7 +72,17 @@ const MaxViolationRecords = 100
 
 // ViolationRecord holds the details of a single policy violation.
 type ViolationRecord struct {
-	// timestamp is when the violation occurred.
+	// id is a per-policy unique identifier allocated by the controller
+	// when the record is first observed. It is stable across re-scrapes
+	// of the same logical violation, so consumers can refer to a single
+	// record by id (for example when correlating with external events).
+	//
+	// Stored as int64 (not uint64) for compatibility with the Kubernetes
+	// field-management machinery used by controller-runtime's test
+	// fixtures; the counter is monotonically increasing and never goes
+	// negative, so the sign bit is never set in practice.
+	ID int64 `json:"id"`
+	// timestamp is when the violation last occurred.
 	Timestamp metav1.Time `json:"timestamp"`
 	// podName is the name of the pod where the violation occurred.
 	PodName string `json:"podName"`
@@ -84,6 +94,16 @@ type ViolationRecord struct {
 	NodeName string `json:"nodeName"`
 	// action is the enforcement action taken (monitor or protect).
 	Action string `json:"action"`
+	// workloadName is the name of the workload that owns the pod, taken
+	// from the pod's first owner reference at the time the record was
+	// first observed. Empty if the pod has no owner reference or could
+	// not be looked up.
+	WorkloadName string `json:"workloadName,omitempty"`
+	// workloadKind is the kind of the workload that owns the pod, taken
+	// from the pod's first owner reference at the time the record was
+	// first observed. Empty if the pod has no owner reference or could
+	// not be looked up.
+	WorkloadKind string `json:"workloadKind,omitempty"`
 }
 
 type WorkloadPolicyStatus struct {
@@ -111,6 +131,17 @@ type WorkloadPolicyStatus struct {
 	// reconciliation.
 	// +optional
 	ViolationCount int64 `json:"violationCount,omitempty"`
+	// nextViolationID is the next id to assign to a brand-new violation
+	// record for this policy. It is incremented atomically with the
+	// status update that introduces a new record, so a single reconcile
+	// cannot double-allocate. Fresh policies start at 1.
+	//
+	// Stored as int64 (not uint64) for compatibility with the Kubernetes
+	// field-management machinery used by controller-runtime's test
+	// fixtures; the counter is monotonically increasing and never goes
+	// negative, so the sign bit is never set in practice.
+	// +optional
+	NextViolationID int64 `json:"nextViolationID,omitempty"`
 	// violations is the list of the most recent violation records (max MaxViolationRecords).
 	// Oldest entries are dropped when the limit is reached.
 	// +optional
--- a/api/v1alpha1/workloadpolicy_types.go
+++ b/api/v1alpha1/workloadpolicy_types.go
@@ -122,8 +122,14 @@ type WorkloadPolicyStatus struct {
 	NodesTransitioning []string `json:"nodesTransitioning,omitempty"`
 	// phase indicates the current phase of the workload policy.
 	Phase Phase `json:"phase,omitempty"`
-	// violationCount is the total number of violation records,
-	// including those no longer retained in violations.
+	// violationCount is the total number of unique violation records
+	// ever observed for this policy, including those that have already
+	// been trimmed out of Violations. It also doubles as the per-policy
+	// id allocator: when a brand-new record is first added, the
+	// reconciler bumps ViolationCount and stamps the new value onto the
+	// record as its id, all in the same status update. As a result, the
+	// largest id ever allocated for a policy is always equal to
+	// ViolationCount, and re-scraped (deduped) records do not bump it.
 	//
 	// Note: This value is maintained by the reconciler and reflects
 	// its best-effort view of the system. It is not guaranteed to be
@@ -131,17 +137,6 @@ type WorkloadPolicyStatus struct {
 	// reconciliation.
 	// +optional
 	ViolationCount int64 `json:"violationCount,omitempty"`
-	// nextViolationID is the next id to assign to a brand-new violation
-	// record for this policy. It is incremented atomically with the
-	// status update that introduces a new record, so a single reconcile
-	// cannot double-allocate. Fresh policies start at 1.
-	//
-	// Stored as int64 (not uint64) for compatibility with the Kubernetes
-	// field-management machinery used by controller-runtime's test
-	// fixtures; the counter is monotonically increasing and never goes
-	// negative, so the sign bit is never set in practice.
-	// +optional
-	NextViolationID int64 `json:"nextViolationID,omitempty"`
 	// violations is the list of the most recent violation records (max MaxViolationRecords).
 	// Oldest entries are dropped when the limit is reached.
 	// +optional
--- a/internal/controller/workloadpolicystatus_helpers.go
+++ b/internal/controller/workloadpolicystatus_helpers.go
@@ -167,9 +167,8 @@ func violationRecordKeyOf(r v1alpha1.ViolationRecord) violationRecordKey {
 // resolveScrapedViolations dedupes scraped records against the existing list,
 // allocates ids for new records (workload name/kind are already populated
 // by the agent), and refreshes the timestamp/node on matched records. It
-// returns the merged violations list (unmatched scraped records prepended,
-// existing records in their original order with timestamps and nodes
-// refreshed where matched) and the updated ViolationCount.
+// returns the merged violations list sorted by timestamp descending
+// (newest first) and the updated ViolationCount.
 //
 // ViolationCount doubles as the id allocator: every brand-new record is
 // stamped with the post-increment value of ViolationCount, so the largest
@@ -209,9 +208,18 @@ func resolveScrapedViolations(
 		newRecords = append(newRecords, s)
 	}
 
-	// Prepend the freshly allocated records in scrape order, then keep
-	// the existing list in place (with timestamps refreshed for matches).
 	merged := slices.Concat(newRecords, existing)
+
+	slices.SortStableFunc(merged, func(a, b v1alpha1.ViolationRecord) int {
+		switch {
+		case a.Timestamp.Time.After(b.Timestamp.Time):
+			return -1
+		case b.Timestamp.Time.After(a.Timestamp.Time):
+			return 1
+		default:
+			return 0
+		}
+	})
 
 	// Trim tail (oldest entries) to keep the most recent MaxViolationRecords.
 	if len(merged) > v1alpha1.MaxViolationRecords {
--- a/internal/controller/workloadpolicystatus_helpers.go
+++ b/internal/controller/workloadpolicystatus_helpers.go
@@ -164,67 +164,40 @@ func violationRecordKeyOf(r v1alpha1.ViolationRecord) violationRecordKey {
 	}
 }
 
-// resolveScrapedViolations dedupes scraped records against the existing list,
-// allocates ids for new records (workload name/kind are already populated
-// by the agent), and refreshes the timestamp/node on matched records. It
-// returns the merged violations list sorted by timestamp descending
-// (newest first) and the updated ViolationCount.
-//
-// ViolationCount doubles as the id allocator: every brand-new record is
-// stamped with the post-increment value of ViolationCount, so the largest
-// id ever allocated for a policy is always equal to ViolationCount and
-// re-scraped (deduped) records do not bump it. A fresh policy starts at
-// 0 and the first id allocated is 1.
+// resolveScrapedViolations merges scraped records into existing, deduping by
+// key, allocating monotonically increasing ids, and sorting by timestamp.
 func resolveScrapedViolations(
 	existing []v1alpha1.ViolationRecord,
 	scraped []v1alpha1.ViolationRecord,
 	violationCount int64,
 ) ([]v1alpha1.ViolationRecord, int64) {
-	// Index existing records by their dedup key so we can recognize a
-	// re-scraped record on the first try.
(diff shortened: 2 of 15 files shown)


api/v1alpha1/workloadpolicy_types.go
```diff
@@ -72,7 +72,17 @@ const MaxViolationRecords = 100
...
+	// id is a per-policy unique identifier allocated by the controller
+	// when the record is first observed. It is stable across re-scrapes
+	// of the same logical violation, so consumers can refer to a single
+	// record by id (for example when correlating with external events).
+	//
+	// Stored as int64 (not uint64) for compatibility with the Kubernetes
+	// field-management machinery used by controller-runtime's test
+	// fixtures; the counter is monotonically increasing and never goes
+	// negative, so the sign bit is never set in practice.
+	ID int64 `json:"id"`
```
kyledong-suse, line 84: Do we want to update the RFC in this PR as it specifies `uint64`?

api/v1alpha1/workloadpolicy_types.go
```diff
@@ -111,6 +131,17 @@ type WorkloadPolicyStatus struct {
...
+	// record for this policy. It is incremented atomically with the
+	// status update that introduces a new record, so a single reconcile
+	// cannot double-allocate. Fresh policies start at 1.
+	//
+	// Stored as int64 (not uint64) for compatibility with the Kubernetes
+	// field-management machinery used by controller-runtime's test
+	// fixtures; the counter is monotonically increasing and never goes
+	// negative, so the sign bit is never set in practice.
+	// +optional
+	NextViolationID int64 `json:"nextViolationID,omitempty"`
```
kyledong-suse: here as well, RFC needs to be updated.

andreaterzolo: i need to take a deeper look at the PR but at a first glance i would expect to reuse `ViolationCount` as ID allocator rather than `NextViolationID`. `ViolationCount` should be already incremented at each violation, if i recall correctly today we increment it after a slice of violations but nothing prevent us to increment it at each new violation. Am i missing something?

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -136,20 +155,138 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
...
+			continue
+		}
+
+		// Brand-new record
+		s.ID = nextID
+		nextID++
+		name, kind := lookupWorkload(ctx, c, log, workloadCache, namespace, s.PodName)
+		s.WorkloadName = name
+		s.WorkloadKind = kind
+		newRecords = append(newRecords, s)
```
kyledong-suse: Do we need to do a dedup for those new record as well?

api/v1alpha1/workloadpolicy_types.go
dottorblaster: Yeah you're right actually, I just went full berserk on the code and didn't think about reusing something we already had. Sorry for this :-)

kyledong-suse (commented):
Overall, LGTM. Just few minor comments.

api/v1alpha1/workloadpolicy_types.go
dottorblaster: I had the same worrying, maybe updating the RFC is best but I would do that in the end when we're sure about the approach

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: The dedup applies only to re-scrapes across ticks to I'd stick with the current code, maybe we could add a stricter comment because the current docstring is a little bit misleading I have to admit

kyledong-suse: Thanks for the explanation.

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -136,20 +154,140 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
...
+			continue
+		}
+
+		// Brand-new record: bump the count, stamp the new value as the
+		// record's id, and look up the pod's first owner reference to
+		// populate the workload fields. The increment happens before the
+		// assignment so a fresh policy (count == 0) gets id 1, not 0.
+		violationCount++
+		s.ID = violationCount
+		name, kind := lookupWorkload(ctx, c, log, workloadCache, namespace, s.PodName)
```
holyspectral: Wouldn't it be easier if agent sends workload information via its grpc API?  It will make sure the workload kind/name extraction behavior is consistent in agent/controller, and avoid the situation where the pod has been deleted when `lookupWorkload()` runs.

holyspectral (commented):
LGTM.  Just a comment.

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: Mega facepalm on my end, this approach is way simpler, thanks

api/v1alpha1/workloadpolicy_types.go
andreaterzolo: actually i don't think we need to update the RFC, IMO it's ok, if we change something from the initial RFC

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -136,20 +144,81 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
...
+	}
+
+	var newRecords []v1alpha1.ViolationRecord
+	for _, s := range scraped {
+		key := violationRecordKeyOf(s)
+		if idx, ok := indexByKey[key]; ok {
+			// Same logical record: refresh the time and node in place.
+			// We keep the original id and workload fields — that is the
+			// whole point of the dedup key.
+			existing[idx].Timestamp = s.Timestamp
```
andreaterzolo, line 183: here we are updating the timestamp but we are not reordering the element in the slice so we don't have a consistent ordering by timestamp. With the current code the last element could have the latest timestamp and could be removed. We should probably reorder the slice at the end of the with `slices.SortStableFunc`

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -110,10 +110,18 @@ func buildPolicyStatus(
...
-	// Merge scraped violations into status: prepend new violations to existing,
-	// then trim to the most recent MaxViolationRecords entries.
-	newStatus.Violations = mergeViolations(wp.Status.Violations, scrapedViolations)
-	newStatus.ViolationCount = wp.Status.ViolationCount + int64(len(scrapedViolations))
+	// Dedupe scraped violations against the existing list, allocate ids for
+	// new records (workload name/kind are already populated by the agent),
+	// and refresh the timestamp/node on matched records. The returned int64
+	// is the updated ViolationCount, which doubles as the id allocator (the
+	// most recently allocated id is always equal to ViolationCount).
+	merged, newCount := resolveScrapedViolations(
```
andreaterzolo:
```suggestion
	newStatus.Violations, newStatus.ViolationCount = resolveScrapedViolations(
```

Andreagit97 (commented):
i'm looking at the tests now

internal/controller/workloadpolicystatus_test.go
```diff
@@ -225,64 +230,272 @@ func makeRecord(i int) v1alpha1.ViolationRecord {
...
-				for i := 1; i < v1alpha1.MaxViolationRecords; i++ {
-					recs[i] = makeRecord(i - 1)
-				}
-				return recs
-			}(),
+// workloadKindDeployment is the API kind string we use as a fixture for
+// owner-reference-based workload resolution tests. Defining it as a constant
+// keeps the linter happy about repeated string literals and reads as a
+// named fixture.
+const workloadKindDeployment = "Deployment"
```
andreaterzolo: we already have a constant for this value in `internal/types/workloadkind/workload.go`

internal/controller/workloadpolicystatus_test.go
```diff
@@ -225,64 +230,272 @@ func makeRecord(i int) v1alpha1.ViolationRecord {
...
+	})
+}
+
+// withID returns a copy of r with the given id.
+func withID(r v1alpha1.ViolationRecord, id int64) v1alpha1.ViolationRecord {
+	r.ID = id
+	return r
+}
+
+func TestBuildPolicyStatusAtomicCounter(t *testing.T) {
```
andreaterzolo: it seems we are already testing this logic in the above tests. `resolveScrapedViolations` seems the right method to test if we want to check the updated `ViolationCount`, why do we need to test the same thing with `buildPolicyStatus`?

internal/controller/workloadpolicystatus_test.go
```diff
@@ -225,64 +230,272 @@ func makeRecord(i int) v1alpha1.ViolationRecord {
...
-				}
-				return recs
-			}(),
+// workloadKindDeployment is the API kind string we use as a fixture for
+// owner-reference-based workload resolution tests. Defining it as a constant
+// keeps the linter happy about repeated string literals and reads as a
+// named fixture.
+const workloadKindDeployment = "Deployment"
+
+func TestResolveScrapedViolations(t *testing.T) {
```
andreaterzolo, line 240: i believe a table driven test would simplify things here

internal/controller/workloadpolicystatus_test.go
```diff
@@ -225,64 +230,272 @@ func makeRecord(i int) v1alpha1.ViolationRecord {
...
+		scraped := []v1alpha1.ViolationRecord{makeRecord(2)}
+		status, err := buildPolicyStatus(wp, nil, scraped)
+		require.NoError(t, err)
+		require.Equal(t, int64(6), status.ViolationCount)
+		require.Len(t, status.Violations, 2)
+		require.Equal(t, int64(6), status.Violations[0].ID)
+	})
+}
+
+func TestBuildPolicyStatusReScrapedKeepsId(t *testing.T) {
```
andreaterzolo: also here it seems we are testing the same thing we tested above with `resolveScrapedViolations`

internal/controller/workloadpolicystatus_test.go
```diff
@@ -309,6 +522,120 @@ func TestWorkloadPolicyViolationCount(t *testing.T) {
 	require.Len(t, status.Violations, v1alpha1.MaxViolationRecords)
 }
 
+// TestConcurrentReconcilesDoNotDoubleAllocate fakes a status-update conflict
+// on the second concurrent reconcile and asserts that the policy never ends
+// up with duplicate ids.
+func TestConcurrentReconcilesDoNotDoubleAllocate(t *testing.T) {
```
andreaterzolo: our reconciler starts every 30 seconds, i don't think we can have concurrent reconciliations, am i missing something?

internal/controller/workloadpolicystatus_test.go
dottorblaster: Fair enough, I didn't think about it :grimacing:

internal/controller/workloadpolicystatus_test.go
dottorblaster: Overly defensive. Removed.

api/v1alpha1/workloadpolicy_types.go
```diff
@@ -84,6 +94,16 @@ type ViolationRecord struct {
 	NodeName string `json:"nodeName"`
 	// action is the enforcement action taken (monitor or protect).
 	Action string `json:"action"`
+	// workloadName is the name of the workload that owns the pod, taken
+	// from the pod's first owner reference at the time the record was
```
andreaterzolo, line 98: this is the downside of these long descriptions, they easily go out-of-sync. These info are not taken `from the pod's first owner reference at the time the record was first observed`

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -136,20 +142,89 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
...
+// by the agent), and refreshes the timestamp/node on matched records. It
+// returns the merged violations list sorted by timestamp descending
+// (newest first) and the updated ViolationCount.
+//
+// ViolationCount doubles as the id allocator: every brand-new record is
+// stamped with the post-increment value of ViolationCount, so the largest
+// id ever allocated for a policy is always equal to ViolationCount and
+// re-scraped (deduped) records do not bump it. A fresh policy starts at
+// 0 and the first id allocated is 1.
+func resolveScrapedViolations(
```
andreaterzolo, line 169: I would change the logic in this way. Some points
* i would deduplicate all the records not just the existing ones. It seems we can get this for free. it would be weird to deduplicate only the existing ones.
* is there a reason to update the nodename `existing[idx].NodeName = s.NodeName`. in the key we have the podName so the node should be always the same, or am i missing something?
* the `violationCount` is the total number of violations we have seen (even if they are duplicated) so we should increment it also on deduplication. This means that the violation IDs won't be consecutive  (0,1,2,...) but i don't think this is a requirement, the important thing is that they are unique.

WDYT?

```go
// resolveScrapedViolations dedupes scraped records against the existing list,
// allocates ids for new records and refreshes the timestamp on matched records.
// It returns the merged violations list sorted by timestamp descending
// (newest first) and the updated ViolationCount.
func resolveScrapedViolations(
	existing []v1alpha1.ViolationRecord,
	scraped []v1alpha1.ViolationRecord,
	violationCount int64,
) ([]v1alpha1.ViolationRecord, int64) {
	// Index existing records by their dedup key so we can recognize a
	// re-scraped record on the first try.
	indexByKey := make(map[violationRecordKey]int, len(existing))
	for i, r := range existing {
		indexByKey[violationRecordKeyOf(r)] = i
	}

	for _, s := range scraped {
		key := violationRecordKeyOf(s)
		if idx, ok := indexByKey[key]; ok {
			// if we have a new violation on the same key we update the timestamp and we increment the counter.
			existing[idx].Timestamp = s.Timestamp
		} else {
			s.ID = violationCount
			existing = append(existing, s)
			indexByKey[key] = len(existing) - 1
		}
		violationCount++
	}

	slices.SortStableFunc(existing, func(a, b v1alpha1.ViolationRecord) int {
		// Sort records in descending order (from most recent to oldest).
		// Comparing b to a (instead of a to b) reverses the natural chronological order.
		return b.Timestamp.Time.Compare(a.Timestamp.Time)
	})

	// Trim tail (oldest entries) to keep the most recent MaxViolationRecords.
	if len(existing) > v1alpha1.MaxViolationRecords {
		existing = existing[:v1alpha1.MaxViolationRecords]
	}

	return existing, violationCount
}
```

api/v1alpha1/workloadpolicy_types.go
dottorblaster: I am pretty much against these comments but I saw them proliferating lately, thanks for letting me know that

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: At first I was dubious about deduplicating for new entries as well but it could be overly optimized and I see now I'm really a minority with that approach :smile: it's actually sound, and it simplifies the code a little bit, so I'll go with that, thanks :+1:

Andreagit97 (approved):

api/v1alpha1/workloadpolicy_types.go
holyspectral: oh I hit this today.  I will just use `int64` instead of following what's in the RFC.

PR #717, opened 2026-07-03
feat(api): add clearance of allowlisted violations

dottorblaster:

Adding `ActiveViolationCount` as per our last RFC, and clearing violations after a binary becomes allowlisted :+1: 

fixes #695

**Checklist**:

- [X] squashed commits into logical changes
- [X] includes documentation
- [X] adds unit tests
- [X] adds or updates e2e tests

### DIFF
--- a/api/v1alpha1/workloadpolicy_types.go
+++ b/api/v1alpha1/workloadpolicy_types.go
@@ -124,19 +124,16 @@ type WorkloadPolicyStatus struct {
 	Phase Phase `json:"phase,omitempty"`
 	// violationCount is the total number of unique violation records
 	// ever observed for this policy, including those that have already
-	// been trimmed out of Violations. It also doubles as the per-policy
-	// id allocator: when a brand-new record is first added, the
-	// reconciler bumps ViolationCount and stamps the new value onto the
-	// record as its id, all in the same status update. As a result, the
-	// largest id ever allocated for a policy is always equal to
-	// ViolationCount, and re-scraped (deduped) records do not bump it.
-	//
-	// Note: This value is maintained by the reconciler and reflects
-	// its best-effort view of the system. It is not guaranteed to be
-	// strongly consistent and may be temporarily outdated depending on
-	// reconciliation.
+	// been trimmed out of Violations or cleared because the executable
+	// was later added to an allowlist. It is not guaranteed to be strongly
+	// consistent and may be temporarily outdated.
 	// +optional
 	ViolationCount int64 `json:"violationCount,omitempty"`
+	// activeViolationCount is the number of currently active (non-cleared)
+	// violation records. It is always equal to len(Violations) and is
+	// updated in the same status write.
+	// +optional
+	ActiveViolationCount int `json:"activeViolationCount,omitempty"`
 	// violations is the list of the most recent violation records (max MaxViolationRecords).
 	// Oldest entries are dropped when the limit is reached.
 	// +optional
@@ -160,6 +157,13 @@ func (s *WorkloadPolicyStatus) AddNodeIssue(nodeName string, issue NodeIssue) {
 			Message: "Maximum number of nodes with issues reached",
 		}
 	}
+}
+
+func (wp *WorkloadPolicy) ClearAllowed() []ViolationRecord {
+	return slices.DeleteFunc(wp.Status.Violations, func(v ViolationRecord) bool {
+		rules := wp.Spec.RulesByContainer[v.ContainerName]
+		return rules != nil && slices.Contains(rules.Executables.Allowed, v.ExecutablePath)
+	})
 }
 
 func (s *WorkloadPolicyStatus) SortTransitioningNodes() {
--- a/internal/controller/workloadpolicystatus_helpers.go
+++ b/internal/controller/workloadpolicystatus_helpers.go
@@ -110,16 +110,19 @@ func buildPolicyStatus(
 	}
 	newStatus.ObservedGeneration = wp.Generation
 
+	existingViolations := wp.ClearAllowed()
+
 	// Dedupe scraped violations against the existing list, allocate ids for
 	// new records (workload name/kind are already populated by the agent),
 	// and refresh the timestamp/node on matched records. The returned int64
 	// is the updated ViolationCount, which doubles as the id allocator (the
 	// most recently allocated id is always equal to ViolationCount).
 	newStatus.Violations, newStatus.ViolationCount = resolveScrapedViolations(
-		wp.Status.Violations,
+		existingViolations,
 		scrapedViolations,
 		wp.Status.ViolationCount,
 	)
+	newStatus.ActiveViolationCount = len(newStatus.Violations)
 	return newStatus, nil
 }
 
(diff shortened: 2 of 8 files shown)


internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -164,6 +167,32 @@ func violationRecordKeyOf(r v1alpha1.ViolationRecord) violationRecordKey {
 	}
 }
 
+func clearAllowedViolations(
```
andreaterzolo: here we are talking about small numbers so i think we can simplify this logic in something like this

```go
func clearAllowedViolations(
	violations []v1alpha1.ViolationRecord,
	rulesByContainer map[string]*v1alpha1.WorkloadPolicyRules,
) []v1alpha1.ViolationRecord {
	return slices.DeleteFunc(violations, func(v v1alpha1.ViolationRecord) bool {
		return slices.Contains(rulesByContainer[v.ContainerName].Executables.Allowed, v.ExecutablePath)
	})
}
```

Moreover, I'm under the impression this method could belong to the `WorkloadPolicyStatus`, WDYT?

internal/controller/workloadpolicystatus_test.go
```diff
@@ -537,3 +538,196 @@ func TestGetViolationsByPolicy(t *testing.T) {
...
+		check      func(t *testing.T, got []v1alpha1.ViolationRecord)
+	}{
+		{
+			name: "drops violations for allowed executable/container pairs",
+			violations: []v1alpha1.ViolationRecord{
+				createViolationRecord("app", "/usr/bin/app"),
+				createViolationRecord("app", "/usr/bin/other"),
+				createViolationRecord("sidecar", "/usr/bin/app"),
+			},
+			rules: createRules,
```
andreaterzolo: the method is only used once, i believe we can directly put the map here
```suggestion
			rules: map[string]*v1alpha1.WorkloadPolicyRules{
		"app": {
			Executables: v1alpha1.WorkloadPolicyExecutables{
				Allowed: []string{"/usr/bin/app", "/bin/sh"},
			},
		},
	},
```

internal/controller/workloadpolicystatus_test.go
```diff
@@ -537,3 +538,196 @@ func TestGetViolationsByPolicy(t *testing.T) {
...
+			check: func(t *testing.T, got []v1alpha1.ViolationRecord) {
```
andreaterzolo: wouldn't be easier to just assert the expected violations instead of having a custom check func?
so something like
```go
expectedViolations: []v1alpha1.ViolationRecord{
				createViolationRecord("app", "/usr/bin/other"),
				createViolationRecord("sidecar", "/usr/bin/app"),
			},
```

internal/controller/workloadpolicystatus_test.go
```diff
@@ -537,3 +538,196 @@ func TestGetViolationsByPolicy(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := clearAllowedViolations(tt.violations, tt.rules)
+			tt.check(t, got)
+		})
+	}
+}
+
+func TestBuildPolicyStatusClearance(t *testing.T) {
```
andreaterzolo: i would probably remove this test and leave only `TestBuildPolicyStatusClearanceFreesCapForNewViolations`, since it doesn't seem to add extra coverage

internal/controller/workloadpolicystatus_test.go
```diff
@@ -537,3 +538,196 @@ func TestGetViolationsByPolicy(t *testing.T) {
...
+			"cleared executable should not appear in violations")
+	}
+	found := false
+	for _, v := range status.Violations {
+		if v.ExecutablePath == "/usr/bin/new-exe" {
+			found = true
+			break
+		}
+	}
+	require.True(t, found, "new violation should fit into the freed slot")
```
andreaterzolo: I would do something like this
```suggestion
	require.Contains(t, status.Violations, newViolation)
	require.NotContains(t, status.Violations, removedViolation)
```

Andreagit97 (commented):
Thank you!

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: Good idea, but I would still keep it pure instead of making it something like a mutating method :+1:

andreaterzolo: got it, do you see any particular advantage in not mutating the slice inside the method?

BTW, probably we can simplify even further using the whole workloadpolicy as method receiver raither than just the status.
```go
// existingViolations := wp.Status.ClearAllowed(wp.Spec.RulesByContainer)
existingViolations := wp.ClearAllowed()
```

Not related to this PR but i believe we should turn `buildPolicyStatus` into a method of the WP

dottorblaster: That's exactly the reason :grimacing:

Since we are going to use this function inside `buildPolicyStatus` I wanted to keep a pure layer, do all the computation and just swap everything in just one place, potentially in a method as you said. I think it makes sense, I will turn `buildPolicyStatus` into a method itself :+1:

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -110,16 +110,19 @@ func buildPolicyStatus(
 	}
 	newStatus.ObservedGeneration = wp.Generation
 
+	existingViolations := wp.Status.ClearAllowed(wp.Spec.RulesByContainer)
+
```
holyspectral: Maybe we could move the `ClearAllowed()` call to AFTER `resolveScrapedViolations()`, so new violations won't have a chance to enter `newStatus.Violations` if a `RulesByContainer` exists to prevent it.

api/v1alpha1/workloadpolicy_types.go
```diff
@@ -124,19 +124,16 @@ type WorkloadPolicyStatus struct {
...
-	//
-	// Note: This value is maintained by the reconciler and reflects
-	// its best-effort view of the system. It is not guaranteed to be
-	// strongly consistent and may be temporarily outdated depending on
-	// reconciliation.
+	// been trimmed out of Violations or cleared because the executable
+	// was later added to an allowlist. It is not guaranteed to be strongly
+	// consistent and may be temporarily outdated.
 	// +optional
 	ViolationCount int64 `json:"violationCount,omitempty"`
```
holyspectral, line 131: Ah I see.  I didn't get that you want to keep both `ViolationCount` and `ActiveViolationCount` in the RFC.  I will change my code accordingly.

holyspectral (commented):
Thanks for working on this.  I'd recommend a small fix to address the temporary status inconsistency when `rulesByContainer` is changed.

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: Done in [5e5627e](https://github.com/rancher-sandbox/runtime-enforcer/pull/717/commits/5e5627e5c858b95c4ee1662c4e0f865eceb9a3b1) @Andreagit97, for the `buildPolicyStatus` refactor I would like to do that in another PR to keep it atomic

internal/controller/workloadpolicystatus_helpers.go
dottorblaster: I'd like to tackle this in the same PR in which I'll tackle the refactor of `buildPolicyStatus` because I feel like the two things are very much intertwined :smile:

Andreagit97 (approved):

PR #718, opened 2026-07-03
feat: add go benchmarks for NRI callbacks

### LINKED ISSUE #584: Add go benchmarks for NRI callbacks and validate them in CI
Since we decided to keep the NRI approach in place, it would make sense to add go benchmarks to monitor/simulate the performance of NRI callbacks.

## Acceptance criteria
* add go benchmarks on the 3 NRI callbacks we use: Synchronize/StartContainer/RemoveContainer
* add a CI check to compare the current values to expected ones (could be a bash script or similar)

kyledong-suse:

Add go benchmarks for NRI callbacks and validate them in CI

fixes #584 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,10 @@ jobs:
       - name: test kubectl plugin build for all platforms
         run: make kubectl-plugin-cross
       - run: make test
+      # TODO: replace static threshold check with github-action-benchmark.
+      # See https://github.com/benchmark-action/github-action-benchmark
+      - name: Check NRI callback benchmarks
+        run: make check-nri-benchmarks
       - name: Upload unit-tests coverage to Codecov
         uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
         env:
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,10 @@ controller: generate-ebpf fmt ## Build manager binary.
 test-bpf: generate-ebpf ## Run bpf tests.
 	go test -v ./internal/bpf -count=1 -exec "sudo -E"
 
+.PHONY: check-nri-benchmarks
+check-nri-benchmarks: generate-ebpf ## Validate NRI callback benchmarks against thresholds.
+	hack/ci/check-nri-benchmarks.sh
+
 .PHONY: agent
 agent: generate-ebpf fmt ## Build agent binary.
 	CGO_ENABLED=0 GOOS=linux go build -o bin/agent ./cmd/agent
--- a/hack/ci/check-nri-benchmarks.sh
+++ b/hack/ci/check-nri-benchmarks.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+THRESHOLDS_FILE="${THRESHOLDS_FILE:-${SCRIPT_DIR}/nri-benchmark-thresholds.txt}"
+BENCH_PKG="${BENCH_PKG:-./internal/nri/}"
+BENCH_COUNT="${BENCH_COUNT:-1}"
+BENCH_TOLERANCE_PERCENT="${BENCH_TOLERANCE_PERCENT:-20}"
+UPDATE_BASELINE="${UPDATE_BASELINE:-false}"
+
+if [[ "${GITHUB_ACTIONS:-}" != "true" ]]; then
+  if [[ "$UPDATE_BASELINE" == "true" ]]; then
+    echo "NRI benchmark baseline update runs in CI only" >&2
+    exit 1
+  fi
+  echo "skipping NRI benchmark check (runs in CI only)"
+  exit 0
+fi
+
+bench_names=(
+  BenchmarkPluginSynchronize
+  BenchmarkPluginStartContainer
+  BenchmarkPluginRemoveContainer
+)
+
+run_benchmarks() {
+  set +e
+  local bench_stdout go_test_exit_code
+  bench_stdout="$(go test \
+    -bench='^BenchmarkPlugin(Synchronize|StartContainer|RemoveContainer)$' \
+    -benchmem \
+    -count="$BENCH_COUNT" \
+    -run='^$' \
+    -exec "sudo -n -E" \
+    "$BENCH_PKG")"
+  go_test_exit_code=$?
+  set -e
+
+  mapfile -t bench_output <<<"$bench_stdout"
+
+  declare -gA observed_ns_per_op=()
+  local pending_bench=""
+  record_result() {
+    local name="${1%-*}" ns="$2"
+    if [[ -z "${observed_ns_per_op[$name]:-}" || "$ns" -gt "${observed_ns_per_op[$name]}" ]]; then
+      observed_ns_per_op["$name"]="$ns"
+    fi
+  }
+
+  for line in "${bench_output[@]}"; do
+    if [[ "$line" =~ ^(Benchmark[^[:space:]]+)[[:space:]]+[0-9]+[[:space:]]+([0-9]+)[[:space:]]+ns/op ]]; then
+      record_result "${BASH_REMATCH[1]}" "${BASH_REMATCH[2]}"
+      pending_bench=""
+    elif [[ "$line" =~ ^(Benchmark[^[:space:]]+) ]]; then
+      pending_bench="${BASH_REMATCH[1]}"
+    elif [[ -n "$pending_bench" && "$line" =~ ^[[:space:]]*[0-9]+[[:space:]]+([0-9]+)[[:space:]]+ns/op ]]; then
+      record_result "$pending_bench" "${BASH_REMATCH[1]}"
+      pending_bench=""
+    fi
+  done
+
+  if (( go_test_exit_code != 0 )); then
+    echo "go test failed with exit code ${go_test_exit_code}" >&2
+    echo >&2
+    echo "raw benchmark output:" >&2
+    printf '%s\n' "${bench_output[@]}" >&2
+    exit "$go_test_exit_code"
+  fi
+}
+
+write_baseline_file() {
+  local commit captured
+  commit="$(git -C "$REPO_ROOT" rev-parse HEAD)"
+  captured="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+
+  {
+    echo "# NRI callback benchmark baseline (ns/op)."
+    echo "# commit: ${commit}"
+    echo "# captured: ${captured}"
+    echo "# tolerance: ±${BENCH_TOLERANCE_PERCENT}% — update after intentional performance changes (CI captures commit and datetime)."
+    for bench_name in "${bench_names[@]}"; do
+      local observed_ns="${observed_ns_per_op[$bench_name]:-}"
+      if [[ -z "$observed_ns" ]]; then
+        echo "missing benchmark result for ${bench_name}" >&2
+        echo >&2
+        echo "raw benchmark output:" >&2
+        printf '%s\n' "${bench_output[@]}" >&2
+        exit 1
+      fi
+      echo "${bench_name} ${observed_ns}"
+    done
+  } >"$THRESHOLDS_FILE"
+
+  echo "wrote baseline to ${THRESHOLDS_FILE}"
+  cat "$THRESHOLDS_FILE"
+}
+
+read_baseline_file() {
+  declare -gA baseline_ns_per_op=()
+  while IFS= read -r line; do
+    [[ -z "$line" || "$line" =~ ^[[:space:]]*# ]] && continue
+    read -r bench_name baseline_ns <<<"$line"
+    baseline_ns_per_op["$bench_name"]="$baseline_ns"
+  done <"$THRESHOLDS_FILE"
+}
+
+# With BENCH_COUNT=1 (default), we keep the single ns/op sample with max ns/op. 
+# This is because max/min/median are identical. With BENCH_COUNT > 1, prefer median
+# for regression detection. One run per PR is enough since CI runs this on every PR.
+check_against_baseline() {
+  local failed=0
+  local min_factor=$((100 - BENCH_TOLERANCE_PERCENT))
+  local max_factor=$((100 + BENCH_TOLERANCE_PERCENT))
+
+  for bench_name in "${bench_names[@]}"; do
+    local baseline_ns="${baseline_ns_per_op[$bench_name]:-}"
+    local observed_ns="${observed_ns_per_op[$bench_name]:-}"
+
+    if [[ -z "$baseline_ns" ]]; then
+      echo "missing baseline for ${bench_name} in ${THRESHOLDS_FILE}" >&2
+      failed=1
+      continue
+    fi
+
+    if [[ -z "$observed_ns" ]]; then
+      echo "missing benchmark result for ${bench_name}" >&2
+      failed=1
+      continue
+    fi
+
+    local min_ns=$((baseline_ns * min_factor / 100))
+    local max_ns=$((baseline_ns * max_factor / 100))
+
+    if (( observed_ns < min_ns || observed_ns > max_ns )); then
+      echo "benchmark ${bench_name} outside ±${BENCH_TOLERANCE_PERCENT}% of baseline: observed ${observed_ns} ns/op, baseline ${baseline_ns} ns/op (allowed ${min_ns}-${max_ns} ns/op)" >&2
+      failed=1
+      continue
+    fi
+
+    echo "benchmark ${bench_name}: ${observed_ns} ns/op (baseline ${baseline_ns} ns/op, ±${BENCH_TOLERANCE_PERCENT}% → ${min_ns}-${max_ns} ns/op)"
+  done
+
+  if (( failed != 0 )); then
+    echo >&2
+    echo "raw benchmark output:" >&2
+    printf '%s\n' "${bench_output[@]}" >&2
+    echo >&2
+    echo "update ${THRESHOLDS_FILE} after intentional performance changes" >&2
+    exit 1
+  fi
+}
+
+if [[ ! -f "$THRESHOLDS_FILE" && "$UPDATE_BASELINE" != "true" ]]; then
+  echo "thresholds file not found: $THRESHOLDS_FILE" >&2
+  echo "create ${THRESHOLDS_FILE} from a CI benchmark run" >&2
+  exit 2
+fi
+
+cd "$REPO_ROOT"
+run_benchmarks
+
+if [[ "$UPDATE_BASELINE" == "true" ]]; then
+  write_baseline_file
+  exit 0
+fi
+
+read_baseline_file
+check_against_baseline
--- a/hack/ci/nri-benchmark-thresholds.txt
+++ b/hack/ci/nri-benchmark-thresholds.txt
@@ -1,0 +1,4 @@
+# NRI callback benchmark baseline (ns/op).
+BenchmarkPluginSynchronize 16629755
+BenchmarkPluginStartContainer 38394
+BenchmarkPluginRemoveContainer 21070
--- a/internal/nri/plugin_bench_test.go
+++ b/internal/nri/plugin_bench_test.go
@@ -1,0 +1,213 @@
+package nri
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/containerd/nri/pkg/api"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/bpf"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/cgroups"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/resolver"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/types/policymode"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	benchPodCount         = 200
+	benchContainersPerPod = 2
+	benchMapMaxEntries    = 1 << 16
+)
+
+type benchMaps struct {
+	cgTracker *ebpf.Map
+	cgToPol   *ebpf.Map
+}
+
+func setupBenchMaps(b *testing.B) *benchMaps {
+	b.Helper()
+
+	require.Equal(b, 0, os.Geteuid())
+
+	newMap := func() *ebpf.Map {
+		m, mapErr := ebpf.NewMap(&ebpf.MapSpec{
+			Type:       ebpf.Hash,
+			KeySize:    8, // cgroupID
+			ValueSize:  8, // trackerID / policyID
+			MaxEntries: benchMapMaxEntries,
+		})
+		require.NoError(b, mapErr)
+		b.Cleanup(func() { _ = m.Close() })
+		return m
+	}
+
+	return &benchMaps{
+		cgTracker: newMap(),
+		cgToPol:   newMap(),
+	}
+}
+
+func newBenchPlugin(b *testing.B, maps *benchMaps) *plugin {
+	b.Helper()
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelDebug,
+	})).With("component", "nri-plugin")
+
+	// The NRI benchmark path never attaches a policy, so these are never invoked.
+	unusedBinaries := func(resolver.PolicyID, []string, bpf.PolicyValuesOperation) error {
+		return errors.New("this function should be unused")
+	}
+	unusedMode := func(resolver.PolicyID, policymode.Mode, bpf.PolicyModeOperation) error {
+		return errors.New("this function should be unused")
+	}
+
+	r, err := resolver.NewResolver(
+		logger,
+		func(cgID uint64, cgroupPath string) error {
+			return bpf.UpdateCgTrackerMap(logger, maps.cgTracker, cgID, cgroupPath)
+		},
+		func(polID resolver.PolicyID, cgroupIDs []resolver.CgroupID, op bpf.CgroupPolicyOperation) error {
+			return bpf.UpdateCgroupPolicy(maps.cgToPol, polID, cgroupIDs, op)
+		},
+		unusedBinaries,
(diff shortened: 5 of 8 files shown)


Makefile
```diff
@@ -108,6 +108,14 @@ controller: generate-ebpf fmt ## Build manager binary.
 test-bpf: generate-ebpf ## Run bpf tests.
 	go test -v ./internal/bpf -count=1 -exec "sudo -E"
 
+.PHONY: bench-nri
+bench-nri: generate-ebpf ## Run NRI callback benchmarks.
+	go test -bench='^BenchmarkPlugin(Synchronize|StartContainer|RemoveContainer)$$' -benchmem -count=1 -run='^$$' ./internal/nri/
```
holyspectral: Maybe a silly question.  Do you not need `sudo` to install bpf progs/maps during the test?

holyspectral (commented):
Just a quick comment for now while I'm looking into the implementation.

Makefile
kyledong-suse: It doesn't need `sudo` to run this `bench-nri`, because the `newBenchPlugin` uses `resolver.NewTestResolver`.  It doesn't touch real bpf progs/maps.

holyspectral: hmm okay.  IMO it's better cover the real bpf progs/maps installation because sometimes the performance issue can happen in the kernel side.  Of course if it's a pretty big change then I'd say let's leave it.  WDYT?

kyledong-suse: I was thinking to use the real bpf progs/maps for the benchmark test, but I also think the mock is fast and stable CI regression detection. I would prefer to keep the mock for now, but let's discuss in tomorrow's daily.

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,107 @@
...
+	benchPodCount         = 10
+	benchContainersPerPod = 3
+)
+
+func newBenchPlugin(b *testing.B) *plugin {
+	b.Helper()
+
+	var cgroupID resolver.CgroupID
+	return &plugin{
+		logger:   slog.New(slog.DiscardHandler),
```
andreaterzolo: actually i would use a real logger, in the end in production we will have a logger for the plugin

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,107 @@
...
+		resolver: resolver.NewTestResolver(b),
+		resolveCgroupID: func(*api.Container) (resolver.CgroupID, string, error) {
```
andreaterzolo: I'm wondering if we should create real cgroups on the host and then call the real function we use in production: https://github.com/rancher-sandbox/runtime-enforcer/blob/01b6e1b2de068c980e4ceb628c07168bbaf3cee0/internal/nri/extract.go#L13
To create a real cgroup we can do something similar to what we do here https://github.com/rancher-sandbox/runtime-enforcer/blob/01b6e1b2de068c980e4ceb628c07168bbaf3cee0/internal/bpf/cgroup_helper_test.go#L38 but we can simplify it because we don't need to run process into this cgroup, we just need to create and cleanup it

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,107 @@
...
+	return pods, containers
+}
+
+func BenchmarkPluginSynchronize(b *testing.B) {
+	pods, containers := buildSyncFixtures(benchPodCount, benchContainersPerPod)
+	ctx := context.Background()
+
+	b.ResetTimer()
+	for range b.N {
+		p := newBenchPlugin(b)
```
andreaterzolo: we should probably stop the timer and start it again like in the other cases

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,107 @@
...
+			})
+		}
+	}
+
+	return pods, containers
+}
+
+func BenchmarkPluginSynchronize(b *testing.B) {
+	pods, containers := buildSyncFixtures(benchPodCount, benchContainersPerPod)
+	ctx := context.Background()
```
andreaterzolo:
```suggestion
	ctx := b.Context()
```

Makefile
andreaterzolo: Taking a closer look i don't think we really need to use our production ebpf prog/maps.
Today the real resolver takes
```go
	resolver, err := resolver.NewResolver(
		logger,
		bpfManager.GetCgroupTrackerUpdateFunc(),
		bpfManager.GetCgroupPolicyUpdateFunc(),
		bpfManager.GetPolicyUpdateBinariesFunc(),
		bpfManager.GetPolicyModeUpdateFunc(),
	)
```

Where each of this function is a method of the bpfManager, but to be honest we use the manager just to obtain the right map

```go
func (m *Manager) GetCgroupTrackerUpdateFunc() func(cgID uint64, cgroupPath string) error {
	return func(cgID uint64, cgroupPath string) error {
		return m.handleErrOnShutdown(updateCgTrackerMap(m.logger, m.objs.CgtrackerMap, cgID, cgroupPath))
	}
}

func updateCgTrackerMap(logger *slog.Logger, cgTrackerMap *ebpf.Map, cgID uint64, cgroupPath string) error {
	// we populate the entry for the cgroup id with itself as tracker id so that the child cgroups
	// can inherit the same tracker id
	if err := cgTrackerMap.Update(&cgID, &cgID, ebpf.UpdateAny); err != nil {
		return fmt.Errorf("failed to update cgroup tracker map for id %d: %w", cgID, err)
	}

	// If the container is not yet started it is impossible to have nested cgroups.
	// In this case, we provide an empty cgroup path since we don't need to walk it.
	if cgroupPath == "" {
		return nil
	}

	// We now walk the cgroup path to find all the child cgroups and map them to the same tracker id.
	// This is useful if the container is already running and has already created child cgroups.
	var walkErr error
	err := filepath.WalkDir(cgroupPath, func(p string, d os.DirEntry, err error) error {
		if err != nil {
			if d == nil {
				return fmt.Errorf("cgrouptracker: failed to walk dir %s: %w", p, err)
			}
			return fs.SkipDir
		}
		if !d.IsDir() {
			return nil
		}

		if p == cgroupPath {
			return nil
		}

		trackedID, err := cgroups.GetCgroupIDFromPath(p)
		if err != nil {
			walkErr = errors.Join(walkErr, fmt.Errorf("failed to read id from '%s': %w", p, err))
			return nil
		}

		// the key here is the child cgroup id we just found
		merr := cgTrackerMap.Update(&trackedID, &cgID, ebpf.UpdateAny)
		if merr != nil {
			walkErr = errors.Join(walkErr, fmt.Errorf("failed to update id (%d) for '%s': %w", trackedID, p, merr))
		}

		logger.Info("added nested cgroup",
			"nested ID", trackedID,
			"parent ID", cgID,
			"nested path", p,
			"parent path", cgroupPath)

		return nil
	})
	if err != nil {
		logger.Warn("failed to run walkdir", "error", err)
	}

	// we just log the error here, as the main update operation could be successful even if some child cgroups failed
	if walkErr != nil {
		logger.Warn("failed to retrieve some of the cgroup ids for some paths", "cgtracker", true, "error", walkErr)
	}
	return nil
}
```

If we change the signature of these methods used by the resolver
```go
func NewResolver(
	logger *slog.Logger,
	cgTrackerUpdateFunc func(cgID uint64, cgroupPath string) error,
	cgroupToPolicyMapUpdateFunc func(polID PolicyID, cgroupIDs []CgroupID, op bpf.CgroupPolicyOperation) error,
	policyUpdateBinariesFunc func(policyID uint64, values []string, op bpf.PolicyValuesOperation) error,
	policyModeUpdateFunc func(policyID uint64, mode policymode.Mode, op bpf.PolicyModeOperation) error,
) (*Resolver, error) {
```

into something that directly takes a map

```go
func NewResolver(
	logger *slog.Logger,
	cgTrackerUpdateFunc func(m *ebpf.Map, cgID uint64, cgroupPath string) error,
	cgroupToPolicyMapUpdateFunc func(m *ebpf.Map, polID PolicyID, cgroupIDs []CgroupID, op bpf.CgroupPolicyOperation) error,
	policyUpdateBinariesFunc func(m *ebpf.Map, policyID uint64, values []string, op bpf.PolicyValuesOperation) error,
	policyModeUpdateFunc func(m *ebpf.Map, policyID uint64, mode policymode.Mode, op bpf.PolicyModeOperation) error,
) (*Resolver, error) {
```

We can directly call `updateCgTrackerMap` with our test map instead of using the manager one.

So the idea would be

```go
func NewTestResolver(t testing.TB) *Resolver {
	t.Helper()

   // create needed ebpf maps

	r, err := NewResolver(
		slog.New(slog.NewJSONHandler(testWriter{t}, nil)),
		UpdateCgTrackerMap(...),
		UpdateCgroupPolicy(...),
		...
	)
	require.NoError(t, err)
	return r
}
```

Maybe i'm missing something but the NRI path reads only cgroup paths and updates ebpf maps. WDYT?

andreaterzolo: The idea here is to mock the input rather than mocking the methods used by the resolver:
* we provide cgroups path created by us that are not real containers
* we provide ebpf maps created by us that are not the ones we use in production 

the resolver shouldn't notice the difference because it should be agnostic

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+func newBenchPlugin(b *testing.B, maps *benchMaps) *plugin {
+	b.Helper()
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+	})).With("component", "nri-plugin")
+
+	// The NRI benchmark path never attaches a policy, so these are never invoked.
+	unusedBinaries := func(resolver.PolicyID, []string, bpf.PolicyValuesOperation) error { return nil }
```
andreaterzolo: I would return an error so that if in the future these methods are called we notice that
```suggestion
	unusedBinaries := func(resolver.PolicyID, []string, bpf.PolicyValuesOperation) error { return errors.New("this function should be unused") }
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,198 @@
...
+		cgToPol:   newMap(),
+	}
+}
+
+func newBenchPlugin(b *testing.B, maps *benchMaps) *plugin {
+	b.Helper()
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+	})).With("component", "nri-plugin")
```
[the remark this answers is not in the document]
kyledong-suse: The script already handles the output, so we can keep info level here.

hack/ci/nri-benchmark-thresholds.txt
```diff
@@ -0,0 +1,5 @@
+# Maximum allowed ns/op for NRI callback benchmarks.
+# Update when intentional performance changes; run `make bench-nri` to capture baselines.
```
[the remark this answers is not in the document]
kyledong-suse: actually I'm going to change it to `run make check-nri-benchmarks`

.github/workflows/ci.yml
```diff
@@ -31,6 +31,8 @@ jobs:
       - name: test kubectl plugin build for all platforms
         run: make kubectl-plugin-cross
       - run: make test
+      - name: Check NRI callback benchmarks
+        run: make check-nri-benchmarks
```
holyspectral, line 37: It doesn't seem to be mentioned anywhere so I'll write it down here.  I think https://github.com/benchmark-action/github-action-benchmark provides exactly what we need to continue monitoring the performance metrics.  Of course this is currently blocked by RST, but it would be interesting when we move to the new org.

hack/ci/nri-benchmark-thresholds.txt
```diff
@@ -0,0 +1,5 @@
+# Maximum allowed ns/op for NRI callback benchmarks.
+# Update when intentional performance changes; run `make check-nri-benchmarks` to capture baselines.
+BenchmarkPluginSynchronize 10000000
+BenchmarkPluginStartContainer 500000
+BenchmarkPluginRemoveContainer 500000
```
holyspectral: Is it possible to generate the baseline using the script here?  For example, 

1. Run the benchmark script and store the result in a txt file like this one with commit ID and datetime. 
2. When the benchmark runs the next time, it checks the txt file and determine the valid range based on a percentage, e.g.,+-10%. 
3. When the benchmark falls out of range, it fails the CI.  Either the code or the benchmark file has to be updated. 

WDYT?

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	b.ResetTimer()
+	for range b.N {
+		b.StopTimer()
+		p := newBenchPlugin(b, maps)
+		b.StartTimer()
+
+		if err := p.StartContainer(ctx, pod, container); err != nil {
+			b.Fatalf("StartContainer: %v", err)
+		}
```
holyspectral, line 190: I think we'd need some comments here about the expected behavior.  I definitely have to dig into code more, but it feels like while we have new bench plugin each time, these `p.StartContainer()` call will actually update the same pod/container into the same map created via `setupBench(b)`.  This doesn't mean it's wrong, but some OS cache magic could make it slightly faster than the real scenario.

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+	container := benchContainer(b, pod, "remove-container")
+	ctx := b.Context()
+
+	b.ResetTimer()
+	for range b.N {
+		b.StopTimer()
+		p := newBenchPlugin(b, maps)
+		if _, err := p.Synchronize(ctx, []*api.PodSandbox{pod}, []*api.Container{container}); err != nil {
+			b.Fatalf("seed Synchronize: %v", err)
+		}
```
holyspectral: Does it make sense to just run `p.StartContainer()` here?

holyspectral (commented):
Generally LGTM.  Some comments about nri-benchmark-thresholds.txt‎ and benchmark cases.

.github/workflows/ci.yml
kyledong-suse: Agreed.`github-action-benchmark` looks like a good fit for tracking trends over time. I propose to revisit once we're in the new org and RST restrictions are out of the way.

internal/nri/plugin_bench_test.go
kyledong-suse: That makes sense. I'll improve the StartContainer and RemoveContainer a bit with a new container per iteration for the benchmark.

hack/ci/nri-benchmark-thresholds.txt
kyledong-suse: That's a good idea, but I'll make the range a bit wider, +- 20%.

kyledong-suse: The test result will be something like
```
benchmark BenchmarkPluginSynchronize: 1566931 ns/op (baseline 1569816 ns/op, ±20% → 1255852-1883779 ns/op)
benchmark BenchmarkPluginStartContainer: 33197 ns/op (baseline 32937 ns/op, ±20% → 26349-39524 ns/op)
benchmark BenchmarkPluginRemoveContainer: 55094 ns/op (baseline 54197 ns/op, ±20% → 43357-65036 ns/op)
```
I'll add the commit ID and datetime in the threshhold file before this PR is ready to merge and squash all commits.

internal/nri/plugin_bench_test.go
holyspectral, line 190: I'd also call `RemoveContainer()` so each time the `StartContainer()` runs in similar condition.

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+		p := newBenchPlugin(b, maps)
+		b.StartTimer()
+
+		if _, err := p.Synchronize(ctx, pods, containers); err != nil {
+			b.Fatalf("Synchronize: %v", err)
+		}
+	}
+}
+
+func BenchmarkPluginStartContainer(b *testing.B) {
```
holyspectral, line 176: If we don't, could we assign a policy to these containers so we cover the part of `r.applyPolicyToPodIfPresent()`?

holyspectral (commented):
Thank you!

.github/workflows/ci.yml
andreaterzolo: maybe you can leave a comment in the file so that we will remember it, you can use a `todo`

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+		cgTracker: newMap(),
+		cgToPol:   newMap(),
+	}
+}
+
+func newBenchPlugin(b *testing.B, maps *benchMaps) *plugin {
+	b.Helper()
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
```
andreaterzolo: I would put the debug here, so that we put ourselves in the worst case
```suggestion
		Level: slog.LevelDebug,
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+	benchContainersPerPod = 3
+	benchMapMaxEntries    = 1 << 16
+)
+
+type benchMaps struct {
+	cgTracker *ebpf.Map
+	cgToPol   *ebpf.Map
+}
+
+func setupBench(b *testing.B) *benchMaps {
```
andreaterzolo:
```suggestion
func setupBenchMaps(b *testing.B) *benchMaps {
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+	b.Helper()
+
+	if os.Geteuid() != 0 {
+		b.Skip("NRI benchmarks with real BPF maps require root privileges")
+	}
```
andreaterzolo: Since all our benchamarks require these privileges i believe we can simply fail instead of skipping the benchmark. I would avoid cases in which we believe the benchmarks are running but in the end we are just skipping them
```suggestion
	require.Equal(os.Geteuid(), 0) 
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	_, err := cgroups.GetCgroupInfo()
+	require.NoError(b, err)
```
andreaterzolo: i don't think we need this here, we already call this when we will create cgroups

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/rancher-sandbox/runtime-enforcer/internal/bpf"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/cgroups"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/resolver"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/types/policymode"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	benchPodCount         = 10
+	benchContainersPerPod = 3
```
andreaterzolo: Maybe we could put ourselves in the worst case. 200*2 seems a good worst case, WDYT?

```suggestion
	benchPodCount         = 200
	benchContainersPerPod = 2
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+		resolver:        r,
+		resolveCgroupID: cgroupFromContainer,
+	}
+}
+
+func createBenchCgroup(b *testing.B, name string) string {
+	b.Helper()
+
+	root := cgroups.GetCgroupResolutionPrefix()
+	relPath := fmt.Sprintf("runtime-enforcer-bench-%s", name)
```
andreaterzolo: just wondering if we should put all test cgroups under a unique directory, so that if by mistake we leave them on the node we can just remove a unique directory

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		if _, err := p.Synchronize(ctx, pods, containers); err != nil {
+			b.Fatalf("Synchronize: %v", err)
+		}
+	}
+}
+
+func BenchmarkPluginStartContainer(b *testing.B) {
+	maps := setupBench(b)
+	pod := testPodSandbox()
```
andreaterzolo, line 178: can we reuse the same helper function in all the tests or is there some reason to use a different func?
```suggestion
	pods, containers := buildSyncFixtures(b, 1, 1)
```

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+	ctx := b.Context()
+
+	b.ResetTimer()
+	for i := range b.N {
+		b.StopTimer()
+		p := newBenchPlugin(b, maps)
+		container := benchContainer(b, pod, fmt.Sprintf("start-container-%d", i))
```
andreaterzolo, line 185: if we can do this outside the for we will save some time, see the comment above

internal/nri/plugin_bench_test.go
andreaterzolo: i believe the resolver is destroyed at each iteration so in theory `StartContainer` is already running in the same condition, am i missing something?

internal/nri/plugin_bench_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		if err := p.StartContainer(ctx, pod, container); err != nil {
+			b.Fatalf("StartContainer: %v", err)
+		}
+	}
+}
+
+func BenchmarkPluginRemoveContainer(b *testing.B) {
+	maps := setupBench(b)
+	pod := testPodSandbox()
```
andreaterzolo, line 196: Same comment of `BenchmarkPluginStartContainer` regarding reusing the same helper method

Andreagit97 (commented):
Thanks the new approach seems powerful! just left some minor comments

hack/ci/check-nri-benchmarks.sh
```diff
@@ -0,0 +1,157 @@
...
+run_benchmarks() {
+  mapfile -t bench_output < <(
+    go test \
+      -bench='^BenchmarkPlugin(Synchronize|StartContainer|RemoveContainer)$' \
+      -benchmem \
+      -count="$BENCH_COUNT" \
+      -run='^$' \
+      -exec "sudo -n -E" \
+      "$BENCH_PKG"
+  )
```
dottorblaster: I think this will swallow the exit code for `go test` because `errexit`/`pipefail` do not apply to process substitution, is it intended?

hack/ci/check-nri-benchmarks.sh
```diff
@@ -0,0 +1,157 @@
...
+read_baseline_file() {
+  declare -gA baseline_ns_per_op=()
+  while IFS= read -r line; do
+    [[ -z "$line" || "$line" =~ ^[[:space:]]*# ]] && continue
+    read -r bench_name baseline_ns <<<"$line"
+    baseline_ns_per_op["$bench_name"]="$baseline_ns"
+  done <"$THRESHOLDS_FILE"
+}
+
+check_against_baseline() {
```
dottorblaster, line 114: I think we have to keep an eye on this function because `observed_ns` keeps the maximum ns/op across counts (`"$ns" -gt "${observed_ns_per_op[$name]}"`). Once you raise `BENCH_COUNT`, taking the worst sample maximizes noise and biases straight toward the upper-bound failure. For regression detection we might want the minimum or the median maybe?

hack/ci/check-nri-benchmarks.sh
```diff
@@ -0,0 +1,157 @@
...
+      -count="$BENCH_COUNT" \
+      -run='^$' \
+      -exec "sudo -n -E" \
+      "$BENCH_PKG"
+  )
+
+  declare -gA observed_ns_per_op=()
+  local pending_bench=""
+  record_result() {
+    local name="${1%-*}" ns="$2"
```
dottorblaster, line 48: Aw yeah regexes

dottorblaster (commented):
A couple things to handle but really great work 👍

internal/nri/plugin_bench_test.go
kyledong-suse: I don't have strong opinion of the worst case, but yeah, it sounds a good worst case.

internal/nri/plugin_bench_test.go
holyspectral: I could be wrong but I was under an impression that each resolver still shares the same ebpf map from `setupBench()`.

internal/nri/plugin_bench_test.go
holyspectral: I think this is a good point and perhaps we should let other benchmark runs at the same condition as well with prefilled `podCache`, `wpState` and something like that.

internal/nri/plugin_bench_test.go
kyledong-suse: That's intentional for start benchmark with unique cgroups per iteration. `buildSyncFixtures` is built once with 200 pods x 2 containers, and it will still reuse the same containers every time.

internal/nri/plugin_bench_test.go
kyledong-suse: Same reason as previous comment.

internal/nri/plugin_bench_test.go
kyledong-suse: Same reason as previous comment.

internal/nri/plugin_bench_test.go
kyledong-suse: After checking the code, I think the resolver is destroyed at each iteration, and the ebpf maps are shared. I might be wrong, but `RemoveContainer` seems wouldn't fully reset everything. From the code perspective, it seems will clear the resolver state and `cgToPol`, but not for `cgTracker`.

internal/nri/plugin_bench_test.go
kyledong-suse: That's a good idea for coverage, but it needs more than labeling the pob, but also working with `policyUpdateBinariesFunc` and `policyModeUpdateFunc`. I'd like to have a separate benchmark function to cover that. If you like, I can create a separate issue to track it.

hack/ci/check-nri-benchmarks.sh
kyledong-suse: Good point, it's not intentional. I'll fix it.

hack/ci/check-nri-benchmarks.sh
kyledong-suse: Good catch. I think currently with `BENCH_COUNT=1` max/min/median are the same, so it’s a no-op today. With another thinking, this benchmark will run in our ci with every PR. So it should be good enough one run per PR as a gate. WDYT?

dottorblaster: I'm fine with it, please add a comment mentioning this thread :slightly_smiling_face:

dottorblaster (approved):
LGTM

PR #719, opened 2026-07-04
feat: acknowledge violations via policy annotations

holyspectral:

This PR provides a way for CRD users to acknowledge violations, so they can triage the violations more easily.  

There are still two items left, but this is generally ready for review. 

1. The label key prefix as discussed [here](https://github.com/rancher-sandbox/runtime-enforcer/issues/696#issuecomment-4897623139).
2. Potential conflict with https://github.com/rancher-sandbox/runtime-enforcer/pull/717.

fixes #696

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


api/v1alpha1/workloadpolicy_types.go
```diff
@@ -120,23 +132,34 @@ type WorkloadPolicyStatus struct {
...
 	Violations []ViolationRecord `json:"violations,omitempty"`
+
+	// acknowledgedViolations is the list of the most recent violation records that are acknowledged
+	// by users (max MaxViolationRecords).
+	// Oldest entries are dropped when the limit is reached.
+	// +optional
+	AcknowledgedViolations []AcknowledgedViolationRecord `json:"acknowledgedViolations,omitempty"`
+
+	// nextViolationID is the ID that is available for the next violation record.
+	NextViolationID int64 `json:"nextViolationID,omitempty"`
```
andreaterzolo: why do we need this `NextViolationID`? we discussed this in one of the previous PR of this serie and `ViolationCount` seemed enough. Is there any particular reason to introduce this one again?

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -92,54 +101,145 @@ func computeWpStatus(
...
+	// If anything goes wrong we can retry in the next reconcile.
+	err = r.Status().Patch(ctx, newPolicy.DeepCopy(), patchBase)
+	if err != nil {
+		return err
+	}
+
+	err = r.Patch(ctx, newPolicy.DeepCopy(), patchBase)
+	if err != nil {
+		return err
+	}
```
dottorblaster, line 244: Maybe we want to encapsulate this into a function to make this coupled update even more visible

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -92,54 +101,145 @@ func computeWpStatus(
...
+// processWorkloadPolicy updates the wp.status and wp.annotation in order to acknowledge a violation.
+// NOTE: agent side ignores annotation changes and status change via predicate.GenerationChangedPredicate{}.
 func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
 	ctx context.Context,
 	wp *v1alpha1.WorkloadPolicy,
 	nodesInfo nodesInfoMap,
 	scrapedViolations []v1alpha1.ViolationRecord,
 ) error {
-	status, err := buildPolicyStatus(wp, nodesInfo, scrapedViolations)
+	patchBase := client.MergeFrom(wp.DeepCopy())
```
dottorblaster, line 217: Why the `MergeFrom`?

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -92,54 +101,145 @@ func computeWpStatus(
...
+	// ViolationCount is exactly the next violation ID.
+	wp.Status.ViolationCount = wp.Status.NextViolationID
+	wp.Status.ActiveViolationCount = int64(len(wp.Status.Violations))
+
+	return nil
 }
 
+// processAcknowledgement handles the acknowledge annotations, remove the annotations in place, and return the updated status.
+//
+//nolint:unparam // keep returning error for code consistency
```
dottorblaster, line 153: I'm not understanding the reason of this `nolint` statement :/

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -22,14 +27,19 @@ func convertToPolicyMode(mode string) pb.PolicyMode {
...
-		TotalNodes: len(nodesInfo),
-	}
+) error {
+	// reset related fields
+	status.NodesWithIssues = nil
+	status.TotalNodes = len(nodesInfo)
+	status.SuccessfulNodes = 0
+	status.FailedNodes = 0
+	status.TransitioningNodes = 0
+	status.NodesTransitioning = nil
```
dottorblaster, line 42: Is this related to this PR and the feature we are trying to implement?

charts/runtime-enforcer/templates/controller/deployment.yaml
```diff
@@ -45,6 +45,32 @@ spec:
...
+        - name: OTEL_EXPORTER_OTLP_INSECURE
+          value: "true"
+        {{- end }}
+        {{- if .Values.telemetry.externalCollector.otelCollectorClientCertificateSecret }}
+        - name: OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE
+          value: /tmp/otel-collector-client-certs/tls.crt
+        - name: OTEL_EXPORTER_OTLP_CLIENT_KEY
+          value: /tmp/otel-collector-client-certs/tls.key
+        {{- end }}
+        {{- end }}
```
dottorblaster: I appreciate this since we are supposed to start using env variables but I would defer this to another PR

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -196,3 +200,33 @@ func (r *WorkloadPolicyStatusSync) getViolationsByPolicy(
...
+		otellog.String("container.name", violation.ContainerName),
+		otellog.String("proc.exepath", violation.ExecutablePath),
+		otellog.String("node.name", violation.NodeName),
+		otellog.String("action", violation.Action),
+		otellog.String("workload.name", violation.WorkloadName),
+		otellog.String("workload.kind", violation.WorkloadKind),
+	)
+
+	r.eventLogger.Emit(ctx, rec)
+}
```
dottorblaster, line 232: Can we just keep this without all the plumbing you added? I'd do that plumbing work in another PR.

dottorblaster (changes requested):
Thanks! This for sure goes towards a nice direction.

A couple items:

- Of course the conflicts with main, I hope it won't be too much of a pain to sync with the current status
- The OTEL plumbing: couldn't we just keep the old plumbing in place? I would like to have all that infrastructural work deferred to another PR. Otherwise we risk losing a little bit of focus on this PR's correctness
- I can see you took a lot of functions that were meant to be pure and turned them into functions that tend to mutate pointer receivers or arguments, is that intentional?

charts/runtime-enforcer/templates/controller/deployment.yaml
```diff
@@ -45,6 +45,32 @@ spec:
         env:
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ quote .Values.kubernetesClusterDomain }}
+        {{- if eq .Values.telemetry.collectorStrategy "default" }}
```
andreaterzolo: should we create a common helper with the agent?

internal/controller/workloadpolicystatus_helpers.go
holyspectral: yeah `MergeFrom` is used to create a patch base for the `Patch()` calls later.  `Patch()` in this feature reduces the chance of conflict and provides a better user experience.

internal/controller/workloadpolicystatus_helpers.go
holyspectral: It's just to allow us to always return a `nil` for error in this function.  When a function always returns nil error, unparam will complain.  :-(

charts/runtime-enforcer/templates/controller/deployment.yaml
holyspectral: Like we discussed in the daily, we currently don't send telemetry from controller and this is what we need in order to emit acknowledge otel log.  I will try to make this a helper function so it's clearer.

internal/controller/workloadpolicystatus_helpers.go
holyspectral: yeah unfortunately I think we have to change this part.  Some detail:

1. `processAcknowledgement()` is by nature to mutate the wp status, because it moves the previous `violations` to `acknowledgedViolations`.  
2. The RFC says `If the id does not match any current violation, the annotation is treated as a no-op (still removed), and a warning OTEL log is emitted so the operator can tell their annotation was stale.`, so it's supposed to only deal with those violations which already exist when the annotation is added, which means it has to deal with the acknowledge flow before new violations are added. 

When 1+2, we had to change `processNodeStatus`, the original `computeWpStatus`, to mutate the existing wp status.

Hope this information helps!

api/v1alpha1/workloadpolicy_types.go
holyspectral: No, sorry I simply missed the conversation as the comment was already resolved when I checked the PR.  I'd suggest that we keep this kind of design change somewhere so it won't be missed again.  

Interestingly I landed at exactly the same conclusion when I write the code. 

```
	// ViolationCount is exactly the next violation ID.
	wp.Status.ViolationCount = wp.Status.NextViolationID
```

Andreagit97 (commented):
As discussed yesterday in the standup i believe we need a little bit a cleanup/refactor on the WP status but this is not the right PR to do that, so it LGTM

Andreagit97 (approved):

dottorblaster (approved):
Time for the followups, let's go

PR #732, opened 2026-07-08
refactor: rework WP status management [2/n]

Andreagit97:

This is part of https://github.com/rancher-sandbox/runtime-enforcer/issues/725. In this PR i moved the processNodeStatus helper to a WP method.
There is a rework of the nodePolicyStatus, mainly renaming.
It seems a lot of code but between tests and generated code the patch is pretty small

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


internal/controller/workloadpolicystatus_sync.go
```diff
@@ -94,9 +94,13 @@ func (r *WorkloadPolicyStatusSync) sync(
 		return err
 	}
 
-	violationsByPolicy := r.getViolationsByPolicy(ctx, clients)
 	nodeStatusByPolicy := r.getNodeStatusByPolicy(ctx, clients, wpList.Items)
 
+	if clients, err = r.agentClientPool.UpdatePool(ctx, r.Client); err != nil {
```
andreaterzolo, line 99: As for the commit description

```
today if we mark a client as stale during the nodestatus scraping we
don't use it for the violation scraping, this is not ideal because we
won't receive any error in the violation scraping even if a node was not
considered.
```

So the best thing is to scrape again all the clients and put the violation error in the whitelist.

I can reproduce these 2 errors at each run and they are due to the `TestRollingUpdate` test. In this test we restart the agent and so when the controller tries to query it, the agent is still not ready. Maybe we should implement a logic that allow us to query the old agent that should be still running during the update, but this is out of scope for this PR

api/v1alpha1/policy_node_status.go
```diff
@@ -1,61 +1,124 @@
...
-	if len(s.NodesTransitioning) == 0 {
-		return
-	}
+func (s *WorkloadPolicyStatus) ProcessPolicyNodeStatus(nodes []PolicyNodeStatus) error {
+	s.NodesWithIssues = nil
+	s.NodesTransitioning = nil
+	s.TotalNodes = len(nodes)
+	s.SuccessfulNodes = 0
+	s.FailedNodes = 0
+	s.TransitioningNodes = 0
```
dottorblaster, line 91: Maybe we could have another method for this reset because it looks like a lot of stuff to track and to remember. Hopefully we won't modify this many times, still I'd like to have a place where we reference all of this in an authoritative way

dottorblaster (approved):
LGTM

PR #745, opened 2026-07-09
refactor: rework WP status management [4/n]

Andreagit97:

This is part of https://github.com/rancher-sandbox/runtime-enforcer/issues/725.
In this PR i moved the acknowledge violation logic inside the v1alpha1 package.
I slightly change the behavior of `AcknowledgeViolationsFromAnnotations` so please take a careful look

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/api/v1alpha1/violation_record.go
+++ b/api/v1alpha1/violation_record.go
@@ -2,6 +2,8 @@ package v1alpha1
 
 import (
 	"slices"
+	"strconv"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -108,4 +110,78 @@ func (s *WorkloadPolicyStatus) MergeScrapedViolations(scraped []ViolationRecord)
 	if len(s.Violations) > MaxViolationRecords {
 		s.Violations = s.Violations[:MaxViolationRecords]
 	}
+}
+
+func (wp *WorkloadPolicy) AcknowledgeViolationsFromAnnotations(now metav1.Time) []AcknowledgedViolationRecord {
+	annotations := wp.GetAnnotations()
+	// No annotations -> no violations to acknowledge
+	if len(annotations) == 0 {
+		return nil
+	}
+
+	type annotationInfo struct {
+		annotationKey string
+		reason        string
+	}
+
+	acknowledges := make(map[int64]annotationInfo, len(annotations))
+
+	for k, reason := range annotations {
+		idStr, found := strings.CutPrefix(k, ViolationAcknowledgePrefix)
+		if !found {
+			continue
+		}
+
+		id, err := strconv.ParseInt(idStr, 10, 64)
+		// we could add a validating webhook that validates the annotation format and
+		// that a corresponding violation with that ID exists.
+		// For now we simplify ignore the annotation if there is no a int after the prefix
+		// and we will leave the annotation untouched.
+		if err != nil {
+			continue
+		}
+
+		acknowledges[id] = annotationInfo{
+			annotationKey: k,
+			reason:        reason,
+		}
+	}
+
+	if len(acknowledges) == 0 {
+		return nil
+	}
+
+	if wp.Status.AcknowledgedViolations == nil {
+		wp.Status.AcknowledgedViolations = make([]AcknowledgedViolationRecord, 0)
+	}
+
+	ackToReturn := make([]AcknowledgedViolationRecord, 0)
+
+	wp.Status.Violations = slices.DeleteFunc(wp.Status.Violations, func(v ViolationRecord) bool {
+		info, ok := acknowledges[v.ID]
+		if !ok {
+			return false
+		}
+		// we remove the annotation from the resource
+		delete(annotations, info.annotationKey)
+
+		newAcknowledgement := AcknowledgedViolationRecord{
+			Violation:      v,
+			Reason:         info.reason,
+			AcknowledgedAt: now,
+		}
+		ackToReturn = append(ackToReturn, newAcknowledgement)
+		wp.Status.AcknowledgedViolations = append(wp.Status.AcknowledgedViolations, newAcknowledgement)
+		return true
+	})
+
+	// we order them so that we always truncate the oldest.
+	slices.SortStableFunc(wp.Status.AcknowledgedViolations, func(a, b AcknowledgedViolationRecord) int {
+		return b.AcknowledgedAt.Time.Compare(a.AcknowledgedAt.Time)
+	})
+
+	if len(wp.Status.AcknowledgedViolations) > MaxViolationRecords {
+		wp.Status.AcknowledgedViolations = wp.Status.AcknowledgedViolations[:MaxViolationRecords]
+	}
+	return ackToReturn
 }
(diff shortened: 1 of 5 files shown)


api/v1alpha1/violation_record.go
```diff
@@ -109,3 +111,77 @@ func (s *WorkloadPolicyStatus) MergeScrapedViolations(scraped []ViolationRecord)
...
+	acknowledges := make(map[int64]annotationInfo, len(annotations))
+
+	for k, reason := range annotations {
+		idStr, found := strings.CutPrefix(k, ViolationAcknowledgePrefix)
+		if !found {
+			continue
+		}
+
+		id, err := strconv.ParseInt(idStr, 10, 64)
+		// we could add a validating webhook that validates the annotation format and
```
andreaterzolo, line 136: now in case of error, we leave the annotation and we simply continue. 
In the future i would use a validating webhook to ensure the annotations are correctly set but for now leaving them untouched seems an acceptable UX. the user will see the annotation still there and will notice the error hopefully.
We could also log a warning but if we go for the webhook probably this is not necessary

api/v1alpha1/violation_record.go
```diff
@@ -109,3 +111,77 @@ func (s *WorkloadPolicyStatus) MergeScrapedViolations(scraped []ViolationRecord)
...
+
+	ackToReturn := make([]AcknowledgedViolationRecord, 0)
+
+	wp.Status.Violations = slices.DeleteFunc(wp.Status.Violations, func(v ViolationRecord) bool {
+		info, ok := acknowledges[v.ID]
+		if !ok {
+			return false
+		}
+		// we remove the annotation from the resource
+		delete(annotations, info.annotationKey)
```
andreaterzolo, line 166: we remove the annotation only if we will acknowledge a violation. If we acknowledge violation n2 but in the status there is no violation n2 the annotation will be left untouched.
I believe we can enforce this as well with a validating webhook but the current UX seems acceptable, WDYT?

holyspectral: hmm this is a hard one.  I agreed that validation webhook would be a good direction to go, but at the same time, I also think we should stick to the original design of RFC to delete a annotation that doesn't have a match ID.  This is because the validation webhook is optional.  Without it, leaving the annotation that doesn't match a record would risk for it to acknowledge an unexpected violation later and it will be hard to troubleshoot.  WDYT?

api/v1alpha1/violation_record.go
```diff
@@ -109,3 +111,77 @@ func (s *WorkloadPolicyStatus) MergeScrapedViolations(scraped []ViolationRecord)
...
+		}
+		ackToReturn = append(ackToReturn, newAcknowledgement)
+		wp.Status.AcknowledgedViolations = append(wp.Status.AcknowledgedViolations, newAcknowledgement)
+		return true
+	})
+
+	// we order them so that we always truncate the oldest.
+	slices.SortStableFunc(wp.Status.AcknowledgedViolations, func(a, b AcknowledgedViolationRecord) int {
+		return b.AcknowledgedAt.Time.Compare(a.AcknowledgedAt.Time)
+	})
```
holyspectral, line 181: yeah this is a great enhancement.  Thanks!

api/v1alpha1/violation_record.go
```diff
@@ -109,3 +111,77 @@ func (s *WorkloadPolicyStatus) MergeScrapedViolations(scraped []ViolationRecord)
 		s.Violations = s.Violations[:MaxViolationRecords]
 	}
 }
+
+func (wp *WorkloadPolicy) AcknowledgeViolationsFromAnnotations(now metav1.Time) []AcknowledgedViolationRecord {
```
holyspectral, line 115: Like we discussed elsewhere, I'd prefer having these functions in the controllers and keep the api package simple. Let's discuss further but I don't think this will block this PR right now as this is still alpha stage.

api/v1alpha1/violation_record.go
```diff
@@ -65,7 +67,7 @@ type violationRecordKey struct {
 	action         string
 }
 
-func violationRecordKeyOf(r ViolationRecord) violationRecordKey {
+func (r ViolationRecord) getKey() violationRecordKey {
```
holyspectral: nit: maybe call it `toKey` or just `Key()` since we don't really `get` the key from somewhere?
```suggestion
func (r ViolationRecord) Key() violationRecordKey {
```

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -123,11 +54,15 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
...
-	err := r.processPolicyStatus(ctx, newPolicy, nodes, scrapedViolations)
+	acknowledged, err := r.processPolicyStatus(newPolicy, nodes, scrapedViolations, metav1.NewTime(time.Now()))
 	if err != nil {
 		return err
 	}
 
+	for _, ack := range acknowledged {
+		r.emitAcknowledgedViolationOtelLog(ctx, ack.Violation, ack.Reason)
+	}
+
```
holyspectral: Sorry this is on me, but could you move this for-loop to after the `err = r.Status().Patch(ctx, newPolicy.DeepCopy(), patchBase)` at Ln77-80?  This way if somehow the acknowledged violations fail to be written to API server, it will not send duplicate otel logs when retrying.  Of course I can do that later if you want to merge this first.

api/v1alpha1/violation_record.go
holyspectral: See my other comments.

api/v1alpha1/violation_record.go
holyspectral: I assumed we're talking about a kubewarden policy.  If we're talking about adding a new check in our own webhook in internal/controller/workloadpolicy_webhook.go or ValidatingAdmissionPolicy in helm charts, then forget about it.  I agree we can skip this part.

api/v1alpha1/violation_record.go
dottorblaster: `Key()` would be good I think

dottorblaster (approved):
Besides the things that @holyspectral highlighted, I think we are in a good spot to merge, thanks!

api/v1alpha1/violation_record.go
andreaterzolo: > If we're talking about adding a new check in our own webhook in internal/controller/workloadpolicy_webhook.go 

Yep i meant this one, sorry

api/v1alpha1/violation_record.go
andreaterzolo: this is now an old branch https://github.com/rancher-sandbox/runtime-enforcer/compare/main...Andreagit97:runtime-enforcer:rework-wp-status-tmp?expand=1 but you can see that as a final result there is only one new public method in the v1alpha1 package, i will open the final PR so that you can see the final result

api/v1alpha1/violation_record.go
andreaterzolo: great point thanks

internal/controller/workloadpolicystatus_helpers.go
andreaterzolo: good point, done :)

PR #753, opened 2026-07-14
fix: prevent eBPF cgroup tracker map exhaustion in NRI benchmarks

dottorblaster:
We are really flooding the CI with logs during the NRI benchmark and it's nearly impossible to understand what's happening :smile: 

In the end we kept debug logs but we fixed a bug on the way.


kyledong-suse (commented):
Thanks for catching it!! Just couple of comments.

1. Could you please extract the duplicated code into a helper function?
2. Could you please update the `nri-benchmark-thresholds.txt` with the latest benchmarks?

```
BenchmarkPluginSynchronize  10944789 ns/op
BenchmarkPluginStartContainer  14950 ns/op
BenchmarkPluginRemoveContainer  8340 ns/op
```

holyspectral (commented):
The change looks good to me.  Thanks!  I wonder what log level you changed to the info tho?

kyledong-suse (approved):
LGTM, thanks!

PR #757, opened 2026-07-16
feat(metrics): export runtime_enforcer_active_violations gauge

kyledong-suse:

Export runtime_enforcer_active_violations as a Prometheus gauge

fixes #698 

**Checklist**:

- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -33,8 +34,10 @@ import (
 	"github.com/rancher-sandbox/runtime-enforcer/internal/customloggers/httpserverlogger"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/events"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/grpcexporter"
+	"github.com/rancher-sandbox/runtime-enforcer/internal/metrics"
 
 	otellog "go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/metric"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -295,26 +298,65 @@ func getMetricsServerOptions(logger *slog.Logger, config *Config) (*certwatcher.
 	return metricsCertWatcher, metricsServerOptions
 }
 
-func setupOtel(ctx context.Context, slogger *slog.Logger, config *Config) (func(context.Context) error, error) {
-	var err error
-	var eventShutdown func(context.Context) error
-	if config.otlpEndpoint != "" {
-		var eventLogger otellog.Logger
-		eventLogger, eventShutdown, err = events.Init(
-			ctx,
-			config.otlpEndpoint,
-			config.otlpCACert,
-			config.otlpClientCert,
-			config.otlpClientKey,
-			config.otlpProtocol,
-		)
-		if err != nil {
-			return nil, fmt.Errorf("failed to initialize otel: %w", err)
-		}
-		config.wpStatusSyncConfig.EventLogger = eventLogger
-		slogger.InfoContext(ctx, "OTLP telemetry enabled", "endpoint", config.otlpEndpoint)
+func setupOtel(
+	ctx context.Context,
+	slogger *slog.Logger,
+	config *Config,
+) (func(context.Context) error, error) {
+	if config.otlpEndpoint == "" {
+		return func(context.Context) error { return nil }, nil
 	}
-	return eventShutdown, nil
+
+	var err error
+	var eventShutdown, metricShutdown func(context.Context) error
+	var eventLogger otellog.Logger
+	var activeViolationsGauge metric.Int64Gauge
+	var shutdowns []func(context.Context) error
+
+	eventLogger, eventShutdown, err = events.Init(
+		ctx,
+		config.otlpEndpoint,
+		config.otlpCACert,
+		config.otlpClientCert,
+		config.otlpClientKey,
+		config.otlpProtocol,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize otel logs: %w", err)
+	}
+	config.wpStatusSyncConfig.EventLogger = eventLogger
+	shutdowns = append(shutdowns, eventShutdown)
+
+	activeViolationsGauge, metricShutdown, err = metrics.Init(
+		ctx,
+		config.otlpEndpoint,
+		config.otlpCACert,
+		config.otlpClientCert,
+		config.otlpClientKey,
+		config.otlpProtocol,
+		config.wpStatusSyncConfig.UpdateInterval,
+	)
+	if err != nil {
+		_ = eventShutdown(ctx)
+		return nil, fmt.Errorf("failed to initialize otel metrics: %w", err)
+	}
+	config.wpStatusSyncConfig.ActiveViolationsGauge = activeViolationsGauge
+	shutdowns = append(shutdowns, metricShutdown)
+
+	slogger.InfoContext(ctx, "OTLP telemetry enabled", "endpoint", config.otlpEndpoint)
+
+	return func(shutdownCtx context.Context) error {
+		var errs error
+		for _, shutdown := range shutdowns {
+			if shutdown == nil {
+				continue
+			}
+			if shutdownErr := shutdown(shutdownCtx); shutdownErr != nil {
+				errs = errors.Join(errs, shutdownErr)
+			}
+		}
+		return errs
+	}, nil
 }
 
 func main() {
@@ -366,9 +408,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	var eventShutdown func(context.Context) error
+	var otelShutdown func(context.Context) error
 
-	eventShutdown, err = setupOtel(ctx, slogger, &config)
+	otelShutdown, err = setupOtel(ctx, slogger, &config)
 	if err != nil {
 		setupLog.Error(err, "failed to setup otel")
 		os.Exit(1)
@@ -404,9 +446,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	if eventShutdown != nil {
-		if err = eventShutdown(ctx); err != nil {
-			slogger.ErrorContext(ctx, "failed to shutdown violation event pipeline", "error", err)
+	if otelShutdown != nil {
+		if err = otelShutdown(ctx); err != nil {
+			slogger.ErrorContext(ctx, "failed to shutdown otel telemetry pipeline", "error", err)
 		}
 	}
 }
--- a/internal/controller/workloadpolicystatus_helpers.go
+++ b/internal/controller/workloadpolicystatus_helpers.go
@@ -15,6 +15,7 @@ import (
 )
 
 // processWorkloadPolicy updates the wp.status and wp.annotation in order to acknowledge a violation.
+// On success, *wp is replaced with the patched policy. On failure, *wp is left unchanged.
 // NOTE: agent side ignores annotation changes and status change via predicate.GenerationChangedPredicate{}.
 func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
 	ctx context.Context,
@@ -56,6 +57,7 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
 	if err != nil {
 		return err
 	}
+	*wp = *newPolicy
 	return nil
 }
 
--- a/internal/controller/workloadpolicystatus_sync.go
+++ b/internal/controller/workloadpolicystatus_sync.go
@@ -10,7 +10,9 @@ import (
 	"github.com/rancher-sandbox/runtime-enforcer/internal/grpcexporter"
 	"github.com/rancher-sandbox/runtime-enforcer/internal/types/loglevel"
 
+	"go.opentelemetry.io/otel/attribute"
 	otellog "go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/metric"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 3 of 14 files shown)


cmd/controller/main.go
```diff
@@ -295,26 +297,66 @@ func getMetricsServerOptions(logger *slog.Logger, config *Config) (*certwatcher.
...
+	shutdowns = append(shutdowns, eventShutdown)
+
+	metricShutdown, err = metrics.Init(
+		ctx,
+		config.otlpEndpoint,
+		config.otlpCACert,
+		config.otlpClientCert,
+		config.otlpClientKey,
+		config.otlpProtocol,
+		mgr.GetClient(),
```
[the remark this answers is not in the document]
kyledong-suse: I have concern about each gauge collection does a live List against the API server. I think it's acceptable if the gauge is ~seconds behind status after a reconcile.

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,152 @@
...
+					want: map[string]int64{"ns1/policy-a": 3},
+				},
+				{
+					sets: []gaugeEntry{{namespace: "ns1", name: "policy-a", count: 7}},
+					want: map[string]int64{"ns1/policy-a": 7},
+				},
+			},
+		},
+		{
+			name: "deleted policy drops out",
```
dottorblaster:
```suggestion
			name: "prune keeps only listed policies",
```

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -86,6 +90,11 @@ func (r *WorkloadPolicyStatusSync) sync(
 
 	if len(wpList.Items) == 0 {
 		r.logger.V(loglevel.VerbosityDebug).Info("No WorkloadPolicies found, retrying later")
+		// Drop any stale gauge series from policies that no longer exist.
+		r.activeViolations.PruneStale(nil)
```
dottorblaster: I understood what you're trying to do but this call is non-obvious. Please also document the `nil` call possibility in the docstring for `PruneStale`

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -46,6 +46,13 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
 		return err
 	}
 
+	// Refresh the gauge from the persisted status.
```
dottorblaster:
```suggestion
	// Refresh the gauge from the status just written.
```

dottorblaster (commented):
Just some this and then we can merge, thank you so much!

internal/metrics/metrics.go
```diff
@@ -0,0 +1,208 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
+	"go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	ActiveViolationsMetricName = "runtime_enforcer_active_violations"
```
andreaterzolo: at the moment we don't need to export this one or am i missing something?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,208 @@
...
+			return nil
+		}),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to register %s gauge: %w", ActiveViolationsMetricName, err)
+	}
+	return nil
+}
+
+func createMetricGRPCExporter(ctx context.Context,
```
andreaterzolo, line 34: These 2 methods are really similar to what we have in the events package `internal/events/events.go` can we reuse the same logic? in the end we just need to return a slice of `opts` where both `otlpmetricgrpc.Option` and `otlploggrpc.Option` implement the same go interface
```go
// Option applies a configuration option to the Exporter.
type Option interface {
	applyGRPCOption(oconf.Config) oconf.Config
}
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,208 @@
...
+	return a.provider.ForceFlush(ctx)
+}
+
+// RegisterGauge registers the runtime_enforcer_active_violations
+// gauge on the given meter.
+func (a *ActiveViolations) RegisterGauge(meter metric.Meter) error {
+	if a == nil {
+		return errors.New("active violations reporter is nil")
+	}
+	_, err := meter.Int64ObservableGauge(
```
andreaterzolo: an `Int64ObservableGauge` is really useful when we scrape something in an asynchronous way, but here we want to update the gauge at each iteration of the controller for each policy so why don't we simply flush the metrics in a synchronous way?

Andreagit97 (commented):
my main doubt is on the synchronous/asynchronous nature of the otel metric

internal/metrics/metrics.go
kyledong-suse: We have to export it because the unit test `package metrics_test` needs to use it.

internal/metrics/metrics.go
kyledong-suse: After checking the definitions, `otlpmetricgrpc.Option` and `otlploggrpc.Option` are different types. Metrics use [applyGRPCOption(oconf.Config)](https://github.com/open-telemetry/opentelemetry-go/blob/exporters/otlp/otlpmetric/otlpmetricgrpc/v1.44.0/exporters/otlp/otlpmetric/otlpmetricgrpc/config.go#L20), logs use [applyOption(config)](https://github.com/open-telemetry/opentelemetry-go/blob/exporters/otlp/otlplog/otlploggrpc/v0.20.0/exporters/otlp/otlplog/otlploggrpc/config.go#L84). So we cannot share one `Option` interface for both log and metric exporters.
I was thinking to add small helpers to trim the endpoint, and the insecure check. But each exporter still needs to build its own opts. And the helpers may need to add unit tests, so I didn't do it.

internal/metrics/metrics.go
kyledong-suse: Agree, move to a sync way with `Int64Gauge`.

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -118,6 +129,11 @@ func (r *WorkloadPolicyStatusSync) sync(
 		}
```
andreaterzolo, line 135: why don't we directly call `gauge.Record` here?
```go
gauge.Record(
	ctx,
	len(wp.ActiveViolations),
	metric.WithAttributes(
		attribute.String("policy.name", wp.Name),
		attribute.String("policy.name", wp.Namespace),
	),
)
```

It's not clear to me why we store the policy names into a map and then we flush them later. More in detail i don't understand the need of the `ActiveViolations` struct. 
If i understand correctly it is enough to call `gauge.Record` and then the otel sdk will periodically flush the gauge to the otel collector. It seems we are implementing our own flushing logic but it's not clear to me why we are doing that

internal/metrics/metrics.go
andreaterzolo: in this repo unit tests always live in the same package of the code, so i think we should change the package name and not export the const

kyledong-suse: After checking most of the packages, I agree that we should keep the same package of the code.
I was coincidentally referring [learning_controller_test.go](https://github.com/rancher-sandbox/runtime-enforcer/blob/main/internal/eventhandler/learning_controller_test.go#L1) which uses the test package :P

internal/controller/workloadpolicystatus_sync.go
kyledong-suse: As we discussed, will `Record` and `Flush` directly in `sync`.

internal/controller/workloadpolicystatus_helpers.go
```diff
@@ -21,13 +21,13 @@ func (r *WorkloadPolicyStatusSync) processWorkloadPolicy(
 	wp *v1alpha1.WorkloadPolicy,
 	nodes []v1alpha1.PolicyNodeStatus,
 	scrapedViolations []v1alpha1.ViolationRecord,
-) error {
+) (int, error) {
 	patchBase := client.MergeFrom(wp.DeepCopy())
 	newPolicy := wp.DeepCopy()
```
andreaterzolo, line 28: i would probably return the newPolicy instead of the activeCount, the caller should only be aware that the returned pointer points to the new policy.

```go
	defer func() {
		// We want to return the updated policy so that we can use the ActiveCount
		*wp = *newPolicy
	}()
```

Moreover returning `0` as activeCount could be dangerous because it means that in case of failure the gauge becomes 0 while we want to keep the old value before the update

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -104,20 +110,36 @@ func (r *WorkloadPolicyStatusSync) sync(
...
+			)
 		}
 	}
 
+	if r.activeViolations != nil {
+		if err = r.activeViolations.Flush(ctx); err != nil {
+			r.logger.Error(err, "failed to export active violations metrics")
+		}
+	}
 	return nil
```
andreaterzolo, line 138: what about something like this (see my previous comment on the return of `processWorkloadPolicy`)

```go
	for _, wp := range wpList.Items {
		policyName := wp.NamespacedName()
		if err = r.processWorkloadPolicy(
			ctx,
			&wp,
			nodeStatusByPolicy[policyName],
			violationsByPolicy[policyName],
		); err != nil {
			r.logger.Error(
				err,
				"failed to process workload policy",
				"policy", policyName,
			)
			continue
		}

		// we update the gauge only in case of success
		r.activeViolations.Gauge.Record(
			ctx,
			int64(wp.Status.ActiveViolationCount),
			metric.WithAttributes(
				attribute.String("policy.name", wp.Name),
				attribute.String("policy.namespace", wp.Namespace),
			),
		)
	}
	return nil
```

If i understand correctly the sdk, we don't need the `Flush` method. You initialize the gauge with `periodicReader := sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithInterval(fallbackExportInterval))` this should mean that the SDK will push for us this gauge to the collector at each `fallbackExportInterval`, and i would say this is enough for our use case

internal/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	activeViolationsMetricName = "runtime_enforcer_active_violations"
+	meterName                  = "active-violations-reporter"
+
+	// fallbackExportInterval is only a safety net for the OTEL PeriodicReader.
```
andreaterzolo: See the other comment, i believe that the periodic reader it's enough for us, we don't need the flushing logic, but please double check it

test/e2e/otel_test.go
```diff
@@ -173,6 +176,69 @@ func getOtelCollectorTest() types.Feature {
...
+	r *resources.Resources,
+	promURL string,
+	policy *v1alpha1.WorkloadPolicy,
+) {
+	t.Helper()
+	const metricName = "runtime_enforcer_active_violations"
+
+	t.Log("validating runtime_enforcer_active_violations gauge matches status.activeViolationCount")
+	var body string
+	require.Eventually(t, func() bool {
```
andreaterzolo, line 194: i'm wondering if we should do exactly what we do for the other metric `runtime_enforcer_violations`. 

We wait for it to be found with the `Eventually` and when we are it is there we just reuse the helpers `assertMetricHasLabel` to assert the labels

```go
				var metricsBody string
				require.Eventually(t, func() bool {
					body, fetchErr := fetchURL(promURL)
					if fetchErr != nil {
						t.Logf("failed to fetch metrics: %v", fetchErr)
						return false
					}
					metricsBody = body
					return strings.Contains(body, "runtime_enforcer_active_violations")
				}, defaultOperationTimeout, 2*time.Second,
					"runtime_enforcer_violations metric should appear on the collector Prometheus endpoint",
				)
```

Here we are doing everything inside Eventually with the limitation that we cannot use explicit assertion but just true/false

internal/controller/workloadpolicystatus_helpers.go
kyledong-suse: >Moreover returning 0 as activeCount could be dangerous because it means that in case of failure the gauge becomes 0 while we want to keep the old value before the update

For returnning 0 as activeCount today won't affect gauge, because on error the code actually will log and skip `Record`. The 0 is actually ignored.

But I agree with you that we can return `newPolicy` instead. However, I'd avoid to assign `newPolicy` if the status patch or annotation patch failure. It's better to assign `*wp = *newPolicy` only after both patches succeed.

internal/controller/workloadpolicystatus_sync.go
kyledong-suse: You're right about the `Flush`. Technically, with only `PeriodicReader`, we don't need `ForceFlush` after each sync because `PeriodicReader` will flush by a timer. The cost is mainly for collect + export + network.
For example, if we set the export interval to 30s, which means the SDK will collect + export every 30s even there is nothing changed. So it'll be extra cost.
For the manually `Flush` on demand, it only exports once per successful sync. However, if sync is too frequent, i.e. every 15s. `Flush` can be more frequent than a 30s reader.
After another check, I think we're doing the WorkloadPolicyStatusSync every 30s via helm chart. So I think it makes sense to change the export interval to 30s to match the sync interval.

andreaterzolo: agreed, in the end doing a flush each 30 secs is equivalent to let the periodicReader do it for us each 30 sec. Maybe instead of using `fallbackExportInterval` in
```go
periodicReader := sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithInterval(fallbackExportInterval))
```
we can reuse the sync interval of the controller so that we are sure this metric is exported with the same frequency

test/e2e/otel_test.go
kyledong-suse: I think for testing the gauge, it's a bit different. We also cares about the value matches status.ActiveViolationCount. So we cannot only wait metric name in the Prometheus is good enough. I already do the label asserts after `Eventually`. So I prefer to keep the test as is.

internal/controller/workloadpolicystatus_sync.go
```diff
@@ -115,6 +120,18 @@ func (r *WorkloadPolicyStatusSync) sync(
...
+			continue
+		}
+
+		if r.activeViolationsGauge != nil {
+			r.activeViolationsGauge.Record(
+				ctx,
+				int64(wp.Status.ActiveViolationCount),
+				metric.WithAttributes(
+					attribute.String("policy.name", wp.Name),
+					attribute.String("k8s.namespace.name", wp.Namespace),
```
andreaterzolo, line 132: is there any particular reason to not call it `policy.namespace`?
```suggestion
					attribute.String("policy.namespace", wp.Namespace),
```

Andreagit97 (commented):
one minor nit and we should be ready to go!

internal/controller/workloadpolicystatus_sync.go
kyledong-suse: The [eventscraper](https://github.com/rancher-sandbox/runtime-enforcer/blob/main/internal/eventscraper/eventscraper.go#L193) uses `"k8s.namespace.name"` which refers to [OTEL k8s semantic convention](https://opentelemetry.io/docs/specs/semconv/resource/k8s/#namespace).
I think we should keep them consistent.

andreaterzolo: got it, thanks!

Andreagit97 (approved):

PR #783, opened 2026-07-29
feat: allow users to directly promote a policy into protect mode

kyledong-suse:

Allow users to directly promote a policy into both monitor or protect mode

fixes #781 

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/api/v1alpha1/keys.go
+++ b/api/v1alpha1/keys.go
@@ -2,12 +2,8 @@ package v1alpha1
 
 const (
 	// ProposalPromoteLabelKey is set on a WorkloadPolicyProposal when it is promoted to a WorkloadPolicy.
-	// Valid values are policymode strings ("monitor", "protect"). The value "true" is accepted as an
-	// alias for "monitor" for backward compatibility.
+	// Valid values are policymode strings ("monitor", "protect").
 	ProposalPromoteLabelKey = "security.rancher.io/promote"
-
-	// ProposalPromoteLabelTrueAlias is the legacy promote label value, treated as "monitor" mode.
-	ProposalPromoteLabelTrueAlias = "true"
 
 	// PolicyPromotedFromLabelKey is set on a WorkloadPolicy when it is created by
 	// promoting a WorkloadPolicyProposal.
--- a/api/v1alpha1/keys.go
+++ b/api/v1alpha1/keys.go
@@ -2,7 +2,12 @@ package v1alpha1
 
 const (
 	// ProposalPromoteLabelKey is set on a WorkloadPolicyProposal when it is promoted to a WorkloadPolicy.
+	// Valid values are policymode strings ("monitor", "protect"). The value "true" is accepted as an
+	// alias for "monitor" for backward compatibility.
 	ProposalPromoteLabelKey = "security.rancher.io/promote"
+
+	// ProposalPromoteLabelTrueAlias is the legacy promote label value, treated as "monitor" mode.
+	ProposalPromoteLabelTrueAlias = "true"
 
 	// PolicyPromotedFromLabelKey is set on a WorkloadPolicy when it is created by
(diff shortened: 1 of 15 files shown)


api/v1alpha1/keys.go
```diff
@@ -2,8 +2,13 @@ package v1alpha1
 
 const (
 	// ProposalPromoteLabelKey is set on a WorkloadPolicyProposal when it is promoted to a WorkloadPolicy.
+	// Valid values are policymode strings ("monitor", "protect"). The value "true" is accepted as an
+	// alias for "monitor" for backward compatibility.
 	ProposalPromoteLabelKey = "security.rancher.io/promote"
 
+	// ProposalPromoteLabelTrueAlias is the legacy promote label value, treated as "monitor" mode.
+	ProposalPromoteLabelTrueAlias = "true"
+
```
dottorblaster: I wouldn't keep the backward compatibility between `true` and  `monitor`

api/v1alpha1/workloadpolicyproposal_types.go
```diff
@@ -67,21 +67,34 @@ func (p *WorkloadPolicyProposal) NamespacedName() string {
...
 	if p == nil {
-		return false
+		return false, ""
+	}
+	val, ok := p.Labels[ProposalPromoteLabelKey]
+	if !ok {
+		return false, ""
+	}
+	switch val {
+	case ProposalPromoteLabelTrueAlias, policymode.MonitorString:
```
dottorblaster:
```suggestion
	case policymode.MonitorString:
```

dottorblaster (changes requested):
A couple changes to make because I don't think we want to be so conservative about the mode label, but everything alright 👍

api/v1alpha1/keys.go
holyspectral: yeah the api version is alpha1, so we're okay to not keep the backward compatibility.

api/v1alpha1/workloadpolicyproposal_types.go
```diff
@@ -67,21 +67,34 @@ func (p *WorkloadPolicyProposal) NamespacedName() string {
...
 		p.SetLabels(map[string]string{})
 	}
-	p.Labels[ProposalPromoteLabelKey] = "true"
+	p.Labels[ProposalPromoteLabelKey] = mode
 }
 
-func (p *WorkloadPolicyProposal) HasPromotionLabel() bool {
+// HasPromotionLabel reports whether the proposal has a valid promotion label and
+// returns the target WorkloadPolicy mode when it does.
+func (p *WorkloadPolicyProposal) HasPromotionLabel() (bool, string) {
```
holyspectral: I'd flip the position of `bool` and `string`.  It's more common to have the boolean value returned as the second value like the line Ln86.

holyspectral (commented):
Could you also add a `ValidatingAdmissionPolicy` or admission webhook, so the label value can be verified?

api/v1alpha1/keys.go
kyledong-suse: It makes sense to me. I'll also update the issue to not keep the backward compatibility.

dottorblaster (approved):
Aw yeah

charts/runtime-enforcer/templates/controller/webhook.yaml
```diff
@@ -59,3 +59,23 @@ webhooks:
...
+  - apiGroups:
+    - security.rancher.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - workloadpolicyproposals
+  sideEffects: None
```
holyspectral, line 81: minor: we probably can have a CEL in `matchConditions` to avoid unnecessary webhook calls.

PR #792, opened 2026-07-30
fix(metrics): separate violation and acknowledge counters

dottorblaster:

The count connector had no conditions, so every emitted log incremented both `runtime_enforcer_violations` and `runtime_enforcer_acknowledge`. Add a body-based condition to each metric so violation and acknowledge logs feed only their own counter, and add the missing `policy.name` / `k8s.namespace.name` attributes to the acknowledge event. E2E asserts the acknowledge counter equals 1 to lock in the split.

### DIFF
--- a/test/e2e/otel_test.go
+++ b/test/e2e/otel_test.go
@@ -362,6 +362,42 @@ func assertMetricHasLabelKey(t *testing.T, body, metricName, labelKey string) {
 		"expected metric %q to have label key %q", metricName, labelKey)
 }
 
+// metricValue returns the value of the first sample line for the given metric
+// carrying the label key=value pair, or false if there is none.
+func metricValue(body, metricName, labelKey, labelValue string) (float64, bool) {
+	expected := fmt.Sprintf(`%s="%s"`, labelKey, labelValue)
+	for line := range strings.SplitSeq(body, "\n") {
+		if !strings.HasPrefix(line, metricName) || !strings.Contains(line, expected) {
+			continue
+		}
+		fields := strings.Fields(line)
+		if len(fields) < 2 {
+			continue
+		}
+		v, err := strconv.ParseFloat(fields[len(fields)-1], 64)
+		if err != nil {
+			continue
+		}
+		return v, true
+	}
+	return 0, false
+}
+
+// assertMetricValue checks that the sample selected by the label key=value pair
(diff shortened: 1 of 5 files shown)


test/e2e/otel_test.go
```diff
@@ -362,6 +362,42 @@ func assertMetricHasLabelKey(t *testing.T, body, metricName, labelKey string) {
...
+// equals the expected value.
+func assertMetricValue(t *testing.T, body, metricName, labelKey, labelValue string, expected float64) {
+	t.Helper()
+
+	v, ok := metricValue(body, metricName, labelKey, labelValue)
+	if !ok {
+		assert.Failf(t, "metric sample not found",
+			"expected metric %q with label %s=%q to be present", metricName, labelKey, labelValue)
+		return
+	}
```
holyspectral: It's just my personal taste but I'd use `assert.Truef(t, ok, ...)` instead.  I'd also include the `body` in the error message.

test/e2e/otel_test.go
```diff
@@ -362,6 +362,42 @@ func assertMetricHasLabelKey(t *testing.T, body, metricName, labelKey string) {
 		"expected metric %q to have label key %q", metricName, labelKey)
 }
 
+// metricValue returns the value of the first sample line for the given metric
+// carrying the label key=value pair, or false if there is none.
+func metricValue(body, metricName, labelKey, labelValue string) (float64, bool) {
```
holyspectral, line 374: Does this function really need to deal with floating point here?  If so, let's keep some examples that this function will receive.

Andreagit97 (approved):

holyspectral (commented):
Thanks!  The change is simple but I have some comments about tests.

test/e2e/otel_test.go
dottorblaster: Switching to `require.Truef` would be even better since it also stops the helper :+1: thanks!

test/e2e/otel_test.go
dottorblaster: The Prometheus exposition format types all sample values as floats :( I updated the comments with some examples

kyledong-suse (approved):
LGTM, thanks!

holyspectral (approved):

PR #807, opened 2026-08-05
refactor: return nil in case of deleted policy during updates

Andreagit97:

In our e2e test it is likely that we will call an update on a deleted policy (https://github.com/rancher-sandbox/runtime-enforcer/actions/runs/30979535824/job/92220790634?pr=805#step:6:9) but more in general this is possible also in production and i wouldn't consider it as an error, it is an expected situation. We could use `warn` verbosity but unfortunately we don't have with `go-logr` so i opted for `info`.
This should also fix transient failures we are seeing in CI https://github.com/rancher-sandbox/runtime-enforcer/actions/runs/30979535824/job/92220790634?pr=805#step:6:9

fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


kyledong-suse (approved):
LGTM, thanks!

PR #823, opened 2026-08-11
feat(api): track when nodes entered their current policy status

### LINKED ISSUE #822: feat: record per-node status transition time ("Since") in WorkloadPolicy status
There's no timestamp for when a node entered its current status. `PolicyStatus` is `{code, message}`, `nodesTransitioning` is a `[]string`, and `successfulNodes` is an `int`; none carry a time. We modelled per-node status with custom structs instead of `[]metav1.Condition`, so we don't get `lastTransitionTime` for free. The agent proto (`PolicyStatus{state, mode, message}`) has no timestamp either, so this has to be stamped controller-side by detecting a code change.

Any transition time has to be *preserved across that rebuild* for nodes whose code is unchanged, which means the recompute needs access to the previous status (the way the violation merge keys off the existing list). Only reset the timestamp when a node's code actually changes.

This is also gated on a product decision: if the "Since" column is dropped from the UX, this ticket is moot. Flagging that up front so we don't build plumbing for a column that gets cut.

We should add a per-node `since` (or a conditions-style `lastTransitionTime`), stamped on code change and carried forward otherwise.

Notes:
- Controller-side only; no agent/proto change (the agent doesn't report transition times).
- `processPolicyNodeStatus()` must take the prior status and carry `since` forward for unchanged codes; stamp `now` only on a code change.
- Scope decision: only nodes-with-issues, or all nodes (ready + transitioning too)? The "all nodes" version is coupled to the Ready-node-names work — you can't attach a per-ready-node `since` if you don't keep the ready names.
- Consider adopting a conditions-style per-node struct (`[]metav1.Condition` or a local `NodeCondition` with `lastTransitionTime`) now, so we don't re-plumb transition tracking later. Bigger change, but it's the idiomatic k8s shape and gives uniform transition handling.

Touchpoints:
- `api/v1alpha1/policy_node_status.go`: add `Since metav1.Time` to `PolicyStatus` (or a new per-node condition type); teach `processPolicyNodeStatus()` to accept the previous status and preserve timestamps across `resetPolicyNodeStatus()`.
- `api/v1alpha1/workloadpolicy_types.go`: `RecomputeStatus()` threads the prior status into the node-status computation (the object already carries the old status, so it's a signature/plumbing change, not a new fetch).
- `api/v1alpha1/policy_node_status_test.go`: cover timestamp preservation on unchanged code and reset on changed code.
- Regenerate CRD manifests + deepcopy, and `docs/crd.adoc` / `docs/crds`.

dottorblaster:

Putting a `Since` field in the status data so users can now since when a policy has been transitioned into a failed state.

Also converted TransitioningNodes into a slice of structs, it might become a little cumbersome to read but at least we are leaving out of the table turning successful policy statuses into full fledged structs, otherwise it would be a footgun both performance-wise and UX-wise (when it comes to users actually reading the CR) 

fixes #822


kyledong-suse (approved):
LGTM, thanks!

PR #832, opened 2026-08-19
feat(api): track per-record occurrences and first-observed timestamp

### LINKED ISSUE #820: feat: enrich ViolationRecord with per-record occurrence count and first-observed timestamp
The Violations tab needs two per-row fields that don't exist today, both on `ViolationRecord`, both computable at merge time with data that already flows through the pipeline.

**Occurrences.** We only expose `status.violationCount`, a policy-level aggregate incremented in `mergeScrapedViolations()` for every scraped record — it can't be decomposed into "how many times did *this* executable/container/pod combo fire". The UI needs a per-row count distinct from that aggregate.

**Age.** `ViolationRecord.timestamp` is refreshed to the *newest* occurrence on every merge (the `if v.Timestamp.After(...)` branch), so it's a last-seen time. There's no first-seen timestamp to compute age from, and `id` is a monotonic counter, not wall-clock, so it can't stand in.

Both are cheap because the agent already buffers one record per violation event (`violationbuf.Buffer.Record()` once per `reportViolation`, 10k ring buffer) and the controller sees every raw event at merge — it just discards the count on dedup and never remembers the first sighting. So this is controller-side only: no agent/proto/BPF change. The existing `Last Occurrence` column keeps deriving from `timestamp`; only `Occurrences` and `Age` are new.

We should add `occurrences` (init `1` on first observation, `++` on each matched duplicate) and `firstObservedTimestamp` (set once at creation, never mutated).

Notes:
- `occurrences`: new records init `1`, matched records `occurrences++`. The merge loop iterates every scraped event, so multiple duplicates in one batch are counted naturally.
- `occurrences` stays distinct from `violationCount` — the latter (per its own doc-comment) includes trimmed/cleared records and isn't strongly consistent, so don't try to make them reconcile exactly.
- `firstObservedTimestamp`: set in the new-record (`else`) branch from the scraped event's own timestamp (the earliest we know of), *not* `time.Now()`; never touch it on a match — that's the whole point vs `timestamp`.
- Mark both `+optional`; `occurrences` gets `+kubebuilder:default=1`.
- No agent/proto/BPF change; `metav1.Time` deepcopy is handled by codegen.
- Related: #764 (violation ID ordering) touches the same record semantics.

Touchpoints:
- `api/v1alpha1/violation_record.go`: add `Occurrences int64` and `FirstObservedTimestamp metav1.Time` to `ViolationRecord`; in `mergeScrapedViolations()` set `occurrences: 1` + `firstObservedTimestamp` in the new-record branch, and `s.Violations[idx].Occurrences++` in the matched branch.
- `api/v1alpha1/violation_record_test.go`: assert per-record counts across multi-batch merges and across the `maxViolationRecords` trim; assert `firstObservedTimestamp` is stamped on creation and stays stable across re-scrapes while `timestamp` advances.
- Regenerate CRD manifests + deepcopy (`make manifests generate`), and `docs/crd.adoc` / `docs/crds`.

dottorblaster:

Since the UI needs them and users repeatedly told us it is cumbersome to identify individual events, we are exposing a counter for occurrences and a timestamp for when the violation was first seen.

fixes #820


api/v1alpha1/violation_record.go
```diff
@@ -24,6 +24,25 @@ type ViolationRecord struct {
 	ID int64 `json:"id"`
 	// timestamp is when the violation last occurred.
 	Timestamp metav1.Time `json:"timestamp"`
```
andreaterzolo: i'm wondering if we should rename this one `LastSeen` or `LastObservedTimestamp`

api/v1alpha1/violation_record_test.go
```diff
@@ -192,6 +222,68 @@ func TestMergeScrapedViolations(t *testing.T) {
...
+	// Per-record counts accumulate across batches and within a single batch.
+	require.Equal(t, int64(2), findByExe("/x").Occurrences)
+	require.Equal(t, int64(3), findByExe("/y").Occurrences)
+	require.Equal(t, int64(1), findByExe("/z").Occurrences)
+
+	// firstObservedTimestamp is stamped with the first sighting and never
+	// updated, while timestamp tracks the newest occurrence.
+	x := findByExe("/x")
+	require.Equal(t, baseTS, x.FirstObservedTimestamp)
+	require.Equal(t, baseTS.Add(time.Minute), x.Timestamp.Time)
```
andreaterzolo: should we move the assertion on Occurrences here?
```go
require.Equal(t, int64(2), x.Occurrences)
```

Andreagit97 (commented):
Thanks! Just a couple of comments

api/v1alpha1/violation_record.go
dottorblaster: It makes sense, I didn't want to break the existing UI but in the end we can always flag it :grimacing:

api/v1alpha1/violation_record_test.go
dottorblaster: Yep

Andreagit97 (approved):

PR #834, opened 2026-08-19
fix: build after `otel/log` v0.21.0 breaking change

dottorblaster:
Just a tiny fix for the new OTel modules switching to a new way of casting/constructing primitives :+1:


Andreagit97 (approved):
Thanks!

